PR #25, opened 2015-11-19
WIP: replace libc with std::os::raw

SimonSapin:
This doesn’t work yet. I’m getting errors like:

```
 expected `*mut libc::types::common::c95::c_void`,
    found `*mut std::os::raw::c_void`
```

I think one of the dependencies is still using libc, but at this point it’s one yak too many to shave for me right now :) Feel free to take this up or disregard it.

### DIFF
--- a/scanner/src/client_gen.rs
+++ b/scanner/src/client_gen.rs
@@ -21,7 +21,7 @@ pub fn generate_client_api<O: Write>(protocol: Protocol, out: &mut O) {
     writeln!(out, "use std::ptr;").unwrap();
     writeln!(out, "use std::sync::Arc;").unwrap();
     writeln!(out, "use std::sync::atomic::{{AtomicBool, Ordering}};").unwrap();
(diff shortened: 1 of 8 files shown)


vberger:
This is very strange that you get this one if you use wayland via glutin. This is code from a version that is more recent than the one used by glutin (0.3 vs 0.2).

I believe the culprit is actually https://github.com/vberger/wayland-window/blob/master/Cargo.toml#L16

I'm going to correct it.

vberger:
I published wayland-window 1.1 fixing this. Is it better ?

SimonSapin:
It’s possible that this crate is not actually built in servo. But because of how Cargo works, even optional dependencies end up in `Cargo.lock` which is where I looked for any occurrence of `libc 0.1`.

vberger:
I'll take some time to figure out if I can get rid of libc or not at some point, but I doubt it's actually possible.

Closing this for now.

PR #37, opened 2015-12-09
client/egl: fix egl_surface_ptr to a more decent state.

vberger:


vberger:
Also include wayland-sys fixes.

Issue #41, opened 2016-01-05
<Missing title> related to ARM breakage


petevine:
Hi, 

I guess you haven't caught up with this [issue](https://github.com/rust-lang/rust/issues/29867) yet :)
Here's a first, coming from an ARM Linux machine using 1.7 nightly:

````
Compiling wayland-client v0.2.1
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/wayland-client-0.2.1/src/core/registry.rs:248:49: 248:58 error: mismatched types:
 expected `*const u8`,
    found `*const i8`
(expected u8,
    found i8) [E0308]
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/wayland-client-0.2.1/src/core/registry.rs:248     let interface_str = unsafe { CStr::from_ptr(interface) };
                                                                                                                                                           ^~~~~~~~~
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/wayland-client-0.2.1/src/core/registry.rs:248:49: 248:58 help: run `rustc --explain E0308` to see a detailed explanation
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/wayland-client-0.2.1/src/core/output/output.rs:256:44: 256:56 error: mismatched types:
 expected `*const u8`,
    found `*const i8`
(expected u8,
    found i8) [E0308]
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/wayland-client-0.2.1/src/core/output/output.rs:256     let manu_str = unsafe { CStr::from_ptr(manufacturer) };
                                                                                                                                                           ^~~~~~~~~~~~
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/wayland-client-0.2.1/src/core/output/output.rs:256:44: 256:56 help: run `rustc --explain E0308` to see a detailed explanation
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/wayland-client-0.2.1/src/core/output/output.rs:257:45: 257:50 error: mismatched types:
 expected `*const u8`,
    found `*const i8`
(expected u8,
    found i8) [E0308]
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/wayland-client-0.2.1/src/core/output/output.rs:257     let model_str = unsafe { CStr::from_ptr(model) };
                                                                                                                                                            ^~~~~
[...]

PR #42, opened 2016-01-06
Move out of libc in profit of std::os::raw

vberger:
cc #41 

might be a breaking change, could be the occasion to squeeze in the few other planned and release `0.6`


vberger:
The fail of build is from https://github.com/rust-lang/rust/issues/30713 and unrelated to this lib.

PR #47, opened 2016-04-29
Update dependencies

mbrubeck:
* Update to crossbeam 0.2
* Allow the latest versions of byteorder, tempfile, and lazy_static
* Update to xml-rs 0.3 and bitflags 0.5/0.6
* Stop tweaking Cargo.toml.  You can include both a version and a path for each dependency.  Local builds will use the path, but builds that fetch the package from crates.io will use the version.


mbrubeck:
oops, I accidentally based this on an old revision.  Rebasing...

vberger:
How urgent is this?
I'm currently in the middle of some pretty big changes in the lib, so I cannot release a version from `master` now at all.
It'd probably be better if I merge this in the `v0.5.x` branch a release a 0.5.11 out of it I think.

vberger:
But your PR includes changes to things that don't exist on `v0.5.x` branch (I have already started changing a lot of things), so it'll need a proper rebase (and update to the examples too).

mbrubeck:
This is not at all urgent; it's fine if these changes aren't published until 0.6.0 is ready.

mbrubeck:
Fixed the example code and added some missing changes in wayland-server.

vberger:
Alright thanks. Didn't know about specifying both a version and a path in Cargo.toml, neat! :smile:

PR #58, opened 2016-07-30
Added desktop-shell extension

Timidger:
Pretty sure I put everything in the right place, if I didn't let me know and I'll fix it. Thanks for working on such a helpful and easy to extend crate!

### DIFF
--- a/protocols/desktop-shell.xml
+++ b/protocols/desktop-shell.xml
@@ -1,0 +1,138 @@
+<protocol name="desktop">
+
+  <interface name="desktop_shell" version="3">
+    <description summary="create desktop widgets and helpers">
+      Traditional user interfaces can rely on this interface to define the
+      foundations of typical desktops. Currently it's possible to set up
+      background, panels and locking surfaces.
+    </description>
+
+    <request name="set_background">
+      <arg name="output" type="object" interface="wl_output"/>
+      <arg name="surface" type="object" interface="wl_surface"/>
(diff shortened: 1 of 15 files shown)


vberger:
Well, everything looks at its place overall yes.

Just, where does this protocol come from? It's not in https://cgit.freedesktop.org/wayland/wayland-protocols/ and I see no copyright header...

Timidger:
I took it from [sway](https://github.com/SirCmpwn/sway) ([added w/ this commit](https://github.com/SirCmpwn/sway/commit/82db2a57a9ab171e77a0765df5dd0b2d92cb4f70)). I am looking to implement background images in my wayland window manager I'm building and I saw sway used this protocol so I decided to use it. I did not realize it's non standard, it does look a bit like a rebranding of [weston-desktop](https://cgit.freedesktop.org/wayland/weston/tree/protocol/weston-desktop-shell.xml).

vberger:
I'm quite uneasy about integrating nonstandart protocols...

Also, I'm currently in a lot of thoughts regarding the design of the lib, and might (once again) change everything in an attempt to better stick of wayland design (which is not easy with rust). If it is successful, It'd probably allow for using 3rd-party protocols without actually needing to patch wayland-client. So, I'd rather let this PR wait until I've figured all that out.

Timidger:
Fair enough, good luck with the redesign

PR #62, opened 2016-09-08
client: method to query protocol errors

### LINKED ISSUE #33: Add a way to query protocol errors
Get inspiration from https://cgit.freedesktop.org/wayland/weston/tree/tests/weston-test-client-helper.c?id=87743e93036c661b7d46a3ebb2a1a16233064004#n719

Will maybe require some fiddling with the scanner, in order to map back the error to its enum.

vberger:
fixes #33

### DIFF
--- a/wayland-client/src/display.rs
+++ b/wayland-client/src/display.rs
@@ -1,9 +1,10 @@
 use std::io;
+use std::ffi::CStr;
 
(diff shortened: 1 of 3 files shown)


vberger:
Actually, I'm not sure about the semantics of wayland API, this needs double-check.

PR #66, opened 2016-09-22
Destructor safety machinery

### LINKED ISSUE #65: Destructor handling and safety
In the current code, there is a big safety hole around destructors:

Wayland Proxies/Resources objects do not have a `Drop` impl. The rationale behind this is that you don't need to keep them around if the reference provided to you in the handler methods is enough, this stays close to the design of the original C library.

However, there *are* destructor events. So for now, once you've called a destructor method or received a destructor event, you *must not* use the object again, or it may completely mess up the internal state of the C library.

This is obviously *unsafe*, but without any unsafe keyword. But I can't find a way to handle this properly which isn't either very cumbersome for the user of the libs, or adds significant overhead.

My ideas for now:

- Stay like this, and document properly that here are hidden unsafeties in the lib
- Don't provide handles to proxy/resources to the user in the callback, but only an ID. They are responsible to keep everything alive, and finding in their collections the appropriate object given its ID to use it. This is still unsafe, because the user *must* destroy the objects when they receive a destructor event, (on the contrary, the current design destroys the C object automatically, but not the rust handle)
- Automatically track liveness of the objects with a "destroyed" flag. This means this flag must be checked at each method call, involving at least two pointer indirection and an atomic read, I think.

vberger:
Fixes #65

### DIFF
--- a/wayland-client/examples/simple_client.rs
+++ b/wayland-client/examples/simple_client.rs
@@ -23,7 +23,7 @@ fn main() {
     };
 
     event_queue.add_handler(MyHandler);
-    let registry = display.get_registry();
+    let registry = display.get_registry().expect("Registry cannot be destroyed!?");
     event_queue.register::<_, MyHandler>(&registry, 0);
 
     event_queue.sync_roundtrip().unwrap();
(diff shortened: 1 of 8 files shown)


vberger:
With this, every request/event method returns a result-like object, that tells you wether the message went through or was discarded because this proxy/resource is actually already destroyed.

This should fix all safety issues around destructors, as they are now automatically handled by the lib according to the protocol.

PR #70, opened 2016-10-08
fix typo for declare_handler! macro

fangyuanziti:
There is a typo in declare_handler! macro. This PR fixed it.


vberger:
Thanks! :smile:

PR #73, opened 2016-11-06
server: add add_socket_* methods of display

### LINKED ISSUE #71: wayland-server: Add ways of adding sockets to Display.
Wayland servers can have sockets added in several ways:
- via `wl_display_add_socket`. Which either: uses a given display name, `WAYLAND_DISPLAY` or `wayland-0`. (In that order)
- via `wl_display_add_socket_auto`. Which scans through `$XDG_RUNTIME_DIR` for an unused `wayland-$d` (where $d is < 32) and binds the first available. 
- via `wl_display_add_socket_fd`. Which accepts a file descriptor. This can be for:
  sockets created and passed in by parent for socket activation, (eg by systemd), 
  or created specifically by the library user for fine grained control.

My initial thoughts:

The first method seems to be the documented behavior of `create_display`. But, I'm not seeing it happen. Although it may not be the best behavior. By which I mean `create_display` should probably only do what `wl_display_create` does (the get event loop is probably fine). 

Perhaps adding `add_socket` methods to `Display` or `EventLoop`. eg:
- `fn add_socket(&mut self, name: Option<&OsStr>) -> Result<()>`
  just calls `wl_display_add_socket`
- `fn add_socket_auto(&mut self) -> Result<String>`
  calls `wl_display_add_socket_auto`. Returns the resulting string. (Which should be utf8, although an `OsString` might be better anyway)
- `fn add_socket_rawfd(&mut self, fd: RawFd) -> Result<()>`
  calls `wl_display_add_socket_fd` with the passed fd.
- `fn add_socket_intofd<F>(&mut self, fd: F) -> Result<()>   where F: IntoRawFd`
  converts fd to a RawFd and calls `add_socket_rawfd`.
  This as a convience method for use with [unix-sockets](https://github.com/rust-lang-nursery/unix-socket).

I might be able to do somethings about this, but I'm a bit busy over the next week or two.

vberger:
Closes #71

@HornetBlack is it good for you?


HornetBlack:
Yep, looks good.  Nice error messages too.

vberger:
Alright!

PR #77, opened 2016-12-19
Make width and height variables

ideasman42:
This helps make it more obvious whats going on.

Note, could wrap lines here - but didn't since the file is already not strictly 99 width max.


vberger:
Thanks, it is indeed much clearer like that! 😄

PR #80, opened 2017-01-09
server: Add Display::init_shm

darkelement:
Just a small enhancement allowing to use wl_display_init_shm function.

### DIFF
--- a/wayland-server/src/display.rs
+++ b/wayland-server/src/display.rs
@@ -162,6 +162,15 @@ impl Display {
             self.ptr
         )};
     }
+
+    /// Create global object providing default implementation of shared memory protocol.
+    pub fn init_shm(&self) {
+        unsafe { ffi_dispatch!(
+            WAYLAND_SERVER_HANDLE,
+            wl_display_init_shm,
+            self.ptr
+        )};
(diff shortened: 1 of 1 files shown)


vberger:
I, thanks for you PR.

I'm sorry to answer like this, but I won't merge this, because `wl_display_init_shm` is actually fundamentally incompatible with the rust structure I've built on top of `libwayland-server`.

Let me explain: to preserve rust safety, I had to build an precise structure with the `user_data` field that the C lib allows to be attached to a `wl_resource`. The safety of the lib depends on this structure to be respected. And the set of helpers for `wl_shm` provided by `libwayland-server` uses `user_data` for it's own implementation details. Meaning it cannot be used with the rest of my crate, or else we'll see direct and unavoidable memory corruption.

This right thing to do is actually to re-implement this helper logic on top of the generic API (which is really what these helpers from the C lib are, in fact), in pure Rust. I've been planning to do that eventually.

Sorry for the trouble, I should have opened an issue about it, so that it as more clear.

vberger:
cc #81

darkelement:
I see. No trouble.

PR #86, opened 2017-02-04
Add a user_data mechanism

vberger:
Very basic mechanism with very low overhead and very small guarantees. The user can they build whatever they want on to of that.


vberger:
I'm just not sure about using `*mut ()`, I'll need to check that.

PR #88, opened 2017-02-19
Add ressource destructor support

vberger:


vberger:
This is a breaking change regarding `wayland-sys` API

PR #92, opened 2017-03-06
Update declare_handler! to use '::std::result::Result'.

Daggerbot:
I extensively use a custom `Result` type (similar to `std::io::Result`), and `use`ing it causes `declare_handler!` to break. I think some others use custom `Result` types as well. Here's a quick fix.


vberger:
Thanks!

Actually, there are a few other macros that may suffer from the same issue, would you mind fixing them as well in this PR, so that they don't get forgotten?

I'm thinking of:

- https://github.com/vberger/wayland-client-rs/blob/master/wayland-client/src/event_queue.rs#L465-L474
- https://github.com/vberger/wayland-client-rs/blob/master/wayland-server/src/event_loop.rs#L492-L500
- https://github.com/vberger/wayland-client-rs/blob/master/wayland-server/src/event_loop.rs#L517-L526

vberger:
Done in 316077d94af542331b42044e58c1e6f606b7067c

PR #94, opened 2017-03-10
Secondary event sources

### LINKED ISSUE #83: Add the capability to add secondary event sources to an event loop
This will make event loops actually worth it server-side.

Secondary sources to handle:

- [x] fds
- [x] timers
- [x] signals
- [ ] idles

vberger:
Adds the capability of secondary event sources:

- [x] FDs
   - [x] with docs
- [x] timers
   - [x] with docs
- [x] OS signals
   - [x] with docs


vberger:
I have no idea what "idle" event sources are, so I won't implement them here.

PR #97, opened 2017-03-16
travis: cargo worksapces are now stable

vberger:


vberger:
... unfortunately, `static_recursion` will only be stable in 1.17, and I don't know of a nicer way to discriminate the "nightly" feature than the current travis.yml.

So, this'll wait 6 weeks, I guess.

vberger:
Obsoleted by #105

PR #98, opened 2017-03-18
User data safety

vberger:
This implements a safety mechanism regarding foreign wayland objects and `user_data`:

When wrapping a wayland object, we try to recognize a magic pointer we stored in its implementation (that we don't use because we have a custom dispatcher). If this pointer is not there, this means this object is not ours, and we act conservatively and don't try to read its user_data or manage it in any way.


vberger:
All this is obviously a breaking change, and will trigger the release of version 0.9.

I'll try to polish the API as much as possible regarding this before merging this PR.

PR #100, opened 2017-03-22
Make client Proxies and server Resources cloneable

vberger:
I think this is sound for managed objects.

Unmanaged objects should be handled with care, though.


vberger:
I think the `Clone` trait is not a proper way after all, because it hides an unsafety : cloning an unmanaged resource is unsafe actually.

PR #101, opened 2017-03-27
Add fmt::Debug trait for RequestResult

rubdos:
Speaks for itself; I wanted to use `debug!("stuff : {:?}")` on a `RequestResult` to have it nicely in my logs. Turns out a fmt::Debug impl was missing.


vberger:
Thanks!

PR #104, opened 2017-04-17
Rework resource_is_registered function

fangyuanziti:
I find the resource_is_registered function do not work as it expected.
* If the resource is registered by EventLoop.register api, the result of calling resource_is_register will be false.
* If the resource is registered by EventLoop.register_with_destructor, the result of calling resource_is_register will be true.

So I make this pull request to fix it. Here are a few things I have done.
* Fix the  resource_is_registered 
```
        ffi_dispatch!(
            WAYLAND_SERVER_HANDLE,
            wl_resource_instance_of,
            resource.ptr(),
            R::interface_ptr(),
            h as *const _ as *const c_void
        )

```
   to 
```
        ffi_dispatch!(
            WAYLAND_SERVER_HANDLE,
            wl_resource_instance_of,
            resource.ptr(),
            R::interface_ptr(),
            &RUST_MANAGED as *const _ as *const _
        )

```
* Re-implement register by calling register_with_destructor because of their extremely similar logic.

### DIFF
--- a/wayland-server/src/event_loop.rs
+++ b/wayland-server/src/event_loop.rs
@@ -116,35 +116,7 @@ impl EventLoopHandle {
         where R: Resource,
               H: Handler<R> + Any + Send + 'static
     {
-        let h = self.handlers[handler_id].downcast_ref::<H>()
-                    .expect("Handler type do not match.");
-        match resource.status() {
-            ::Liveness::Dead => return RegisterStatus::Dead,
-            ::Liveness::Unmanaged => return RegisterStatus::Unmanaged,
-            ::Liveness::Alive => { /* ok, we can continue */ }
-        }
-        unsafe {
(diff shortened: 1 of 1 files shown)


vberger:
Okay, this is so embarrassing. I can't believe I forgot these when I refactored the whole thing...

Thanks a lot for spotting that! 👍

fangyuanziti:
I think more unit test will be helpful. :smile:

vberger:
Well, unit-testing this kind of library is quite difficult actually. It's hard to test a client without a full-featured server running, and hard to test a server without a client connected to it...

Honestly, if you have any idea how I could automate testing of this library, I'm really willing to listen to it.

vberger:
Hmm, wait, I didn't realise the wayland project had this : https://github.com/wayland-project/wayland/tree/master/tests

I'll take inspiration from it.

PR #105, opened 2017-04-17
Test infrastructure

vberger:
Basic test infrastructure for the set of libs.

Client & Server libs need to be tested together, so we cannot really unit-test them. Luckily cargo supports packages composed entirely of tests, so the root of the workspace will host them. 😄 

I'll add some tests before merging.


vberger:
I'll merge this as is. The infra is here, no tests yet, but adding them will be straightforward.

Issue #116, opened 2017-06-14
wl_listener is defined as an opaque pointer


Timidger:
From what I have seen of most examples in C, when constructing a `wl_listener` to listen for signals a struct is created that has a pointer to `wl_notify_func` (which you define) in the `.notify` field. 

However, in the wayland-sys crate in `server.rs`, it's defined as a opaque pointer which means it can't be constructed by this library directly. Is this a bug? 

Here is the common definition I'm finding in the header files: https://people.freedesktop.org/~whot/wayland-doxygen/wayland/Server/structwl__listener.html

vberger:
This is likely an oversight, as this struct is not directly used by wayland-client or wayland-server...

vberger:
Actually, do you only need `wl_listener`, or do you plan to use the whole `wl_signal` machinery?

I've tried to give a go and expose the `wil_signal_init`/`wl_signal_get`/ etc.. functions, but unfortunately these are defined in the headers of the C libs, and designed to be inlined, as such, they are not exposed in the symbols of the C library. Which means we'd need to re-implement them in rust.

Now, some of them (`wl_signal_get` and `wl_signal_emit`) are built in C using the `wl_list_for_each` macro, itself built using the `wl_container_of` macro. But I don't see how we can implement this `wl_container_of` macro in rust in a way that does not invoke UB. It uses `offsetof`, which we do not have in rust, and I don't see how we can implement it.

Timidger:
I have actually somewhat re implemented those types in my library that uses wayland-rs, so it would be nice to have these here upstream. 

In regards to `wl_container_of`, I have a very unsafe implementation that never the less works. It would only work on `#[repr(C)]`   structs, due to the reordering optimization from the Rust compiler. I don't think these are possible to make safe in a general sense (unless you can guarantee the data passed to the macro is  `#[repr(C)]`, which I don't think you can), but it would be nice if those unsafe primitives were still exposed.

vberger:
I believe this can be closed now. Please open a new issue if you encounter any problem with the implementation. :)

PR #120, opened 2017-06-21
sys: add wl_listener and wl_signal

vberger:
cc #116

### DIFF
--- a/wayland-sys/src/server.rs
+++ b/wayland-sys/src/server.rs
@@ -12,7 +12,6 @@ pub enum wl_display { }
 pub enum wl_event_loop { }
 pub enum wl_event_source { }
 pub enum wl_global { }
-pub enum wl_listener { }
 pub enum wl_resource { }
 pub enum wl_shm_buffer { }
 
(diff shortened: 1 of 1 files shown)


vberger:
This is an attempt at implementing the `wl_signal_*` fucntions from wayland-server's headers in rust.

We need to make sure:

- `offset_of!(..)` is not UB, and if it is, can we work around it?
- the functions are correctly implemented, according to the reference impl of wayland-server

vberger:
@Timidger it looks like you are interested in this, your eyes would be welcome!

Timidger:
I appreciate this, this stuff is super helpful for my wlroots wrapper. 

Haven't had a chance to double check the implementation against the reference, but a quick glance the macros at least look right. 

In regards to the safety of `offset_of!`, it _should_ be safe so long as the struct is `#[repr(C)]`. Without that, Rust can reorder the struct ordering as an optimization. 

Either way though this will rely on the internal memory representation of Rust structs. Making them use the C model means that shouldn't change.

vberger:
My wary regarding `offset_of!()` is that we are kind-of dereferencing a null pointer, isn't that UB?

Timidger:
According to [this stackoverflow answer](https://stackoverflow.com/questions/38819569/how-to-get-a-pointer-to-a-containing-struct-from-a-pointer-to-a-member) (which is where I got that particular snippet), this is a simplification on the solution posted by the questioner. The original is a little more explicit and is definitely not UB because he just does straight pointer arithmetic.

You're correct though that this does seem kind of suspect...I understand the logic done though. I played around with it my self, and I think the magic is that we immediantly convert the field to a pointer, so that Rust is smart enough to not actually dereference the pointer ever actually. It is able to see (even with no optimizations turned on) that we aren't asking for the value of the struct at memory address 0, we are asking for the memory address of the field of the struct that resides at memory address 0. That way, we are just doing basic pointer arithmetic.

[Here is me playing around with it on the playground](https://is.gd/okCbIT). Note the cases where UB can be invoked, but with the code as-is it works because we aren't ever actually de referencing null pointers (even though it looks like we are).

vberger:
Okay, in this case I'll merge this and release a new version.

I hope if there is any bug in this, it won't be too hard to hunt... I've tried to follow closely the C code, though, so it should be ok I guess.

Timidger:
I'll be making heavy use if it, so if I notice anything strange I'll be sure to report it here, and try to provide a fix

PR #133, opened 2017-08-18
Update xml-rs and bitflags to latest versions

pedrocr:


vberger:
Thanks!

The upgrade of bitflags does not imply any breaking change regarding the exposed generated flag structs, right?

pedrocr:
Looking at the code that implemented this it seems the only change is really on the macro to define this so the end result after macro expansion should be the exact same thing.

vberger:
Okay, perfect! 👍

PR #140, opened 2017-09-02
Added a workaround when using dlopen on Ubuntu 17.04

### LINKED ISSUE #139: wayland-sys can not find `libwayland-{server, client}.so` on Ubuntu 17.04 when "dlopen" is enabled
For whatever reason Ubuntu 17.04 does not have a `libwayland-server.so` or `libwayland-client.so` symlinked to their versioned counterparts (e.g. `libwayland-server.so.0.1.0` or even `libwayland-server.so.0`) which means wayland-sys cannot find the respective libraries when it uses `libloading` to dynamically load those libraries.

A workaround for users is to add the `libwayland-server.so` symlink. It might also make sense to try to convince the wayland packagers to add the symlink.

But given everything else, perhaps this library should look for `libwayland-server.so.0` instead, or at least try looking for the other more exact versions if it can't find the first one.

mark-blum:
This adds a maintenance overhead of making sure the listed
`libwayland-{client, server}.so` versions are correct for our current
target unfortunately but should also be completely backwards compatible
as it will still try the "bare" version first and only try the others if
it can not find that one.

Fixes #139

### DIFF
--- a/wayland-sys/src/client.rs
+++ b/wayland-sys/src/client.rs
@@ -81,17 +81,30 @@ external_library!(WaylandClient, "wayland-client",
 #[cfg(feature = "dlopen")]
 lazy_static!(
     pub static ref WAYLAND_CLIENT_OPTION: Option<WaylandClient> = {
-        match WaylandClient::open("libwayland-client.so") {
-            Ok(h) => Some(h),
-            Err(::dlib::DlError::NotFound) => None,
-            Err(::dlib::DlError::MissingSymbol(s)) => {
-                if ::std::env::var_os("WAYLAND_RS_DEBUG").is_some() {
-                    // only print debug messages if WAYLAND_RS_DEBUG is set
-                    eprintln!("[wayland-client] Found library libwayland-client.so cannot be used: symbol {} is missing.", s);
+        // This is a workaround for Ubuntu 17.04, which doesn't have a bare symlink
+        // for libwayland-client.so but does have it with the version numbers for
+        // whatever reason.
+        //
+        // We could do some trickery with str slices but that is more trouble
+        // than its worth
+        let versions = ["libwayland-client.so",
+                        "libwayland-client.so.0",
+                        "libwayland-client.so.0.3.0"];
+        let mut ret = None;
+        for ver in &versions {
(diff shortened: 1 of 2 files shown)


wayland-sys/src/client.rs
```diff
@@ -81,17 +81,30 @@ external_library!(WaylandClient, "wayland-client",
...
+        // This is a workaround for Ubuntu 17.04, which doesn't have a bare symlink
+        // for libwayland-client.so but does have it with the version numbers for
+        // whatever reason.
+        //
+        // We could do some trickery with str slices but that is more trouble
+        // than its worth
+        let versions = ["libwayland-client.so",
+                        "libwayland-client.so.0",
+                        "libwayland-client.so.0.3.0"];
+        let mut ret = None;
```
vberger: I don't think this dans with `ret` is necessary. We can just early-return at the first `Ok(_)`.

vberger:
Can you explain where these `0.3.0` and `0.1.0` version number come from? They do not seem to match the actual version number of the libraries... ?

wayland-sys/src/client.rs
mark-blum: Huh, you're right. I figured that I wouldn't be able to do it since that code block isn't really a function but I guess I was wrong

mark-blum:
Those version numbers are what are in my `/usr/lib/` on both my Ubuntu 17.04 and my Arch computers. I'm not completely sure where they come from exactly though since the Arch package that provides this is at `1.14.0-1` and the Ubuntu package that provides this is at `1.12.0-1`.

Regardless I could remove the full version name and just leave the `.so.0` versions since those seem to work everywhere and would require less maintenance, since we would only have to watch for major version changes in the library.

vberger:
Ah, I see, the ABI of the libs is versioned independently from the libs themselves.

> I could remove the full version name and just leave the .so.0 versions since those seem to work everywhere and would require less maintenance

Yes, I'd be in favor of that. Thanks!

mark-blum:
Updated to only checking `.so` and `.so.0`

vberger:
Thanks!

PR #142, opened 2017-09-18
[breaking] Event loop rework

vberger:
This breaks almost everything, largely changing the way the crates work.

We insert a clear boundary between "data" and "implementation" over the
event loop. Solving various "shared state access" issues.

Also, the event loops/event queues are no longer Send, and can store
non-Send data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target
 Cargo.lock
 *.rs.bk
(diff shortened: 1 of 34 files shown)


vberger:
The build failure here is caused by a rustfmt bug, current version doe snot work with workspaces...

PR #159, opened 2017-11-08
Fixes #158 Can now parse KDE server dec protocol

### LINKED ISSUE #158: Wayland scanner cannot parse KDE server decoration protocol
Issue found [here](https://github.com/swaywm/wlroots-rs/pull/12#issuecomment-342574102).

EDIT: Reproduced here:
```
thread 'main' panicked at 'Ill-formed protocol file: Some(Ok(CData(
    Copyright (C) 2015 Martin Gräßlin

    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU Lesser General Public License as published by
    the Free Software Foundation, either version 2.1 of the License, or
    (at your option) any later version.

    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU Lesser General Public License for more details.

    You should have received a copy of the GNU Lesser General Public License
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
  )))', /home/cosm/.cargo/registry/src/github.com-1ecc6299db9ec823/wayland-scanner-0.12.2/src/parse.rs:70:40
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

This might be an issue with xml-rs, because I _think_ the XML should be valid.

Timidger:
Fixes an issue found while parsing the KDE client side decoration protocol where parsing the copyright clause would cause a panic if it contained a CDATA instead of a just the contents of the copyright notice.

This fixes this by properly unwrapping the string from the `CData`

### DIFF
--- a/wayland-scanner/src/parse.rs
+++ b/wayland-scanner/src/parse.rs
@@ -67,7 +67,12 @@ fn parse_protocol<'a, S: Read + 'a>(mut iter: Events<S>) -> Protocol {
                 match &name.local_name[..] {
                     "copyright" => {
                         // parse the copyright
-                        let copyright = extract_from!(iter => XmlEvent::Characters(copyright) => copyright);
+                        let copyright = match iter.next() {
+                            Some(Ok(XmlEvent::Characters(copyright))) |
+                            Some(Ok(XmlEvent::CData(copyright))) => copyright,
+                            e => panic!("Ill-formed protocol file: {:?}", e)
+                        };
+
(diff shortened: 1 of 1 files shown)


vberger:
Thanks, this is the proper fix. 👍 

My only worry now is whether the code should be allowed to allow CDATA anywhere as well...

Anyway, this is good for now.

PR #164, opened 2018-01-30
Add with_idata to EventSources

Drakulix:


wayland-server/src/event_sources.rs
```diff
@@ -50,6 +50,18 @@ impl<ID> FdEventSource<ID> {
         }
     }
 
+    /// Do something with the implementation data without removing it from the `EventLoop`
```
vberger: s/Do something with/Access/g

wayland-server/src/event_sources.rs
```diff
@@ -50,6 +50,18 @@ impl<ID> FdEventSource<ID> {
...
+    pub fn with_idata<F, R>(&mut self, evlh: &mut EventLoopHandle, f: F) -> R
+    where
+        R: 'static,
```
vberger: Why is this `R: 'static` needed? (applies to all 4 functions)

Drakulix: I did not want the user to return something bound to the lifetime of idata, which would likely be a reference to idata. But of course you could move something with a lifetime into the closure and try to return it.

vberger: I *think* the Rust rules won't allow the lifetime of `R` to be bound to the lifetimes of either argument....

But even if it did, it would mutably borrow `self` and the event loop handle, preventing the user to do anything bad, wouldn't it?

vberger:
You can ignore the coveralls issue, it looks like there is an issue with the travis build unrelated to this PR.

wayland-server/src/event_sources.rs
Drakulix: Yeah sounds like it. I am probably just paranoid.

Drakulix:
Alright, I fixed both your comments and added a new commit.

The last commit moves `with_idata` and `remove` into an `EventSource` trait. I understand if you don't want to merge that now (then just drop it / cherry-pick the other), because thats a breaking change, as it might break imports.

I actually have a use-case for that last change in fireplace however. So if you could potentially queue that for the next major release, that would be nice.

vberger:
Well, next release will be major then.

Could you just add a line in the changelog, in the "unreleased" section?

Drakulix:
Done

vberger:
Thanks!

PR #165, opened 2018-02-25
[WIP] wayland-rs-0.20.0

vberger:
Currently WIP, complete rework of the libs to significantly improve the APIs and reduce the unsafe code use, as well as pave the road for a full-rust implementation of the protocol.

**TODO:**

- **commons:**
  - [x] API documentation
  - [x] Crate documentation

- **scanner:**
  - [x] client code generation
  - [x] server code generation

- **client:**
  - [x] object implementation & dispatching
  - [x] connection initialization
  - [x] multiple queues handling (non-send support)
  - [x] convenience code for handling globals
  - [x] wayland-egl
  - [x] wayland-cursor
  - [x] API documentation
  - [x] Crate documentation

- **server:**
  - [x] object implementation & dispatching
  - [x] non-send implementation support
  - [x] server initialization
  - [x] client tracking
  - [x] multiple event queues & event sources
  - [x] API documentation
  - [x] Crate documentation

PR #167, opened 2018-04-15
Make wayland_server::create_event_loop public

myfreeweb:
It is useful for compositors:

https://github.com/myfreeweb/weston-rs/blob/8222ffab43a41a54ef107454558ca495cbe4d009/weston-rs/src/display.rs#L32-L34

### DIFF
--- a/wayland-server/src/lib.rs
+++ b/wayland-server/src/lib.rs
@@ -105,8 +105,8 @@ extern crate wayland_sys;
 
 pub use client::Client;
 pub use display::{create_display, Display};
-pub use event_loop::{resource_is_registered, EventLoop, EventLoopHandle, Global, GlobalCallback,
-                     RegisterStatus, State, StateProxy, StateToken};
+pub use event_loop::{resource_is_registered, create_event_loop, EventLoop, EventLoopHandle, Global,
+                     GlobalCallback, RegisterStatus, State, StateProxy, StateToken};
 pub use generated::interfaces as protocol_interfaces;
 pub use generated::server as protocol;
 use wayland_sys::common::{wl_argument, wl_interface};


myfreeweb:
hm, I'm actually creating the display myself. I'm not sure why I have my own `wl_display` wrapper... oh, probably because I want to create new `Display` references from a raw `*mut wl_display`.

myfreeweb:
Oh. Actually. I still need this. I have the Rust side implementing an API that Weston calls into.

But maybe for `EventLoopHandle` instead of `EventLoop`

vberger:
If you need it only for `EventLoopHandle` (registering & implementing new resources, I guess ?), then it is possible that the incoming 0.20 rework (see #165 ) will solve your issue, as it largely decouples resources and implementing them from the `EventLoop` type, and `EventLoopHandle` is completely removed. This PR is almost ready, and I'm just checking the few potentials missing bits by working on https://github.com/Smithay/smithay/pull/88 using it.

Overall, this reworks does break a lot of things, but should make many things easier to integrate and handle I believe. Maybe if you summarize the precise uses you need `EventLoop` and `EventLoopHandle` for, I can see if some adaptations to wayland-rs 0.20 should be done accordingly?

myfreeweb:
> registering & implementing new resources, I guess

No. I just need to call `add_fd_event_source` in a Rust impl that's called from C (libweston).

I just realized though, I can actually use `user_data` on libweston's compositor object to pass the Rust event loop object through the C world!...

myfreeweb:
[Done](https://github.com/myfreeweb/weston-rs/commit/5223e65bd0b6f2659f972d59f18a84a55f7479e2).

I guess I was also slightly confused by the lack of a `run` method on your `Display`, but running the `EventLoop` works fine.

PR #169, opened 2018-05-08
client: check libs in `Display::connect_to_env`

b-r-u:
Check if the client libraries are available in
`Display::connect_to_env` before accessing `WAYLAND_CLIENT_HANDLE`.

This should help with https://github.com/tomaka/winit/issues/504.


vberger:
Ah, this is a regrettable oversight... Thanks for spotting this!

PR #170, opened 2018-05-27
[WIP] [breaking] Pure rust implementation of the protocol

vberger:
This is a mega-WIP-PR tracking the rust implementation of the protocol in wayland-rs. Once this is finished and landed, the `native_lib` flag will actually be meaningful.

**General ideas of the implementation::**

- a lot of the message serialization logic is common to client and server, and will live in wayland-commons
- from an API point of view, activating the `native_lib` feature will only add a few methods to access the C pointers of the underlying objects. As a result, any dependency that works without this flag should continue working with.
- client-side, for now the `egl` and `cursor` features depend on `native_lib`. While the `cursor` feature could be implemented in pure rust, it is unlikely the egl feature could easily, unless some radical changes are accepted upstream in [wayland-egl](https://github.com/wayland-project/wayland/tree/master/egl), regarding the backend API.
- all tests must pass both with and without the `native_lib` feature (of course)

**Progress:**

- [x] commons:
  - [x] message serialization and deserialization logic
  - [x] object map abstraction
  - [x] socket abstraction
- [x] scanner:
  - [x] types and traits generation to encode the protocol
  - [x] API generation of the objects
  - [x] `RequestsTrait` generation for client side
- [x] client:
  - [x] connection logic to find the server socket (must match the C libs rules wrt environment variables)
  - [x] event queue & message dispatching logic
  - [x] display implementation, `delete_id` and protocol errors
  - [x] blocking `dispatch`
  - [x] ~~& `prepare_read` API~~ : The `prepare_read` API is not essential for now, it still works while currently mocked though slightly suboptimal. This can be improved later and will not block this PR.
- [x] server:
  - [x] general event loop & sources logic
  - [x] wayland socket listening logic
  - [x] wayland messages dispatching logic
  - [x] globals management logic
  - [x] completed API


vberger:
Well, here we go :tada:

mitchmindtree:
This looks like a seriously impressive amount of work @vberger, epic stuff! Looking forward to having a closer look next time I get a chance. :heart:

vberger:
You can expect a blog post about it soon-ish ;)

PR #171, opened 2018-06-02
protocols: make c_interfaces public

myfreeweb:
Other protocols can reference e.g. xdg_popup_interface.

### DIFF
--- a/wayland-protocols/src/protocol_macro.rs
+++ b/wayland-protocols/src/protocol_macro.rs
@@ -27,6 +27,7 @@ macro_rules! wayland_protocol(
                     pub(crate) use wayland_sys as sys;
                     pub(crate) use wayland_client::protocol::{$($import),*};
                     include!(concat!(env!("OUT_DIR"), "/", $name, "_c_client_api.rs"));
+                    pub use super::c_interfaces;
                 }
             }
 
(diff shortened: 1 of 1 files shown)


vberger:
The C `*const wl_interface` are already exposed through the `c_interface()` associated function of the `Interface` trait: `<XdgPopup as Interface>::c_interface()`.

Do you actually need more than that? If so, what is your use-case? Overall all this is more or less supposed to be an implementation detail...

vberger:
Ah right, I forgot about this one.

In this case, I think it' be good rework the protocol macro slightly, to something like:

```rust
#[cfg(feature = "client")]
pub use self::generated::client;

#[cfg(feature = "server")]
pub use self::generated::server;

#[cfg(feature = "native_lib")]
pub use self::generated::c_interfaces;

#[cfg(feature = "native_lib")]
mod generated {
    pub mod c_interfaces {
        // ...
    }

    #[cfg(feature = "client")]
    pub mod client {
        // the contents of previous generated::client::c_api
    }

    #[cfg(feature = "server")]
    pub mod server {
        // the contents of previous generated::server::c_api
    }
}
```

as actually there is no use to have two versions of `c_interfaces`, given it's exactly the same client-side and server-side.

If you don't want to do it right now, I'll probably integrate it as par of #170 at some point.

PR #172, opened 2018-06-03
client: make make_display public

myfreeweb:
To allow using wayland-rs with `wl_display` objects received from e.g. UI toolkits.

(My use case: setting up a GTK window with a custom shell protocol instead of xdg-shell.)

### DIFF
--- a/wayland-client/src/display.rs
+++ b/wayland-client/src/display.rs
@@ -70,7 +70,11 @@ pub struct Display {
 
 impl Display {
     #[cfg(feature = "native_lib")]
-    unsafe fn make_display(ptr: *mut wl_display) -> Result<(Display, EventQueue), ConnectError> {
+    /// Wrap an existing wayland connection (e.g. from a UI toolkit) into Rust wrappers.
+    ///
+    /// On success, you are given the `Display` object as well as the main `EventQueue` hosting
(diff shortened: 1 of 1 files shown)


vberger:
`make_display` as it is implemented right now makes some pretty strong assumptions: it takes ownership of the `*mut wl_display` pointer, and also assumes full ownership over the display event loop (event loops are not threadsafe and barely synchronized..).

As such, if you use it from a pointer to a display owned by GTK, and whose event loop is managed by GTK, this may likely cause wrong behavior.

If you just need to implement a specific protocol, but let GTK manage the event loop, creating a `Proxy<WlDisplay>` using `Proxy::from_c_ptr(..)` should be enough to retrieve the registry and the global you need.

If you need full access to the display/event_loop, while still sharing it with GTK, I think we'll need to design an other method specifically for that. So, could you tell me more about how this interacts with GTK, and the relevant APIs you are using?

vberger:
Okay, I see.

I this case, for best correctness, I think what should instead be exposed is a function that, given the `*mut wl_display`, creates a new wrapper for it and a new event queue, that will be for use by the rust application. As well a internally mark the fact that we don't own that display (and as such should not disconnect it on drop).

I'll wrap something along those lines.

PR #173, opened 2018-06-03
Improvements to C interop client-side

vberger:
- Allows the use of a foreign `wl_display`
- Expose generated C interfaces for protocols

Fixes #171 
Fixes #172 

@myfreeweb Can you confirm this meets your needs?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- [client] Allow the creation of a `Display` from a foreign `wl_display`
+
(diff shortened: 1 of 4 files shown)


myfreeweb:
It should! But the protocols crate with the `server` feature does not compile, because everything in that c interfaces thing references `wayland_client` stuff

vberger:
Right, I overlooked this part. Now it should be good.

myfreeweb:
Yeah, everything works now! Thanks.

vberger:
Good, I'll make a release with that.

PR #174, opened 2018-06-10
scanner: Fix '*mut_' typo (add space)

myfreeweb:
Was erroring with "expected mut or const in raw pointer type"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


vberger:
Thanks! Could you just add a line to the changelog listing this change?

Out of curiosity, which protocol are you using, that triggered this (which was apparently not triggered by any of the protocols currently in wayland-protocols)?

vberger:
Interesting! Thanks!

PR #175, opened 2018-07-04
Added as_ptr to Display and EventQueue

Timidger:
This is necessary in Way Cooler to set up the client event queue to respond after the glib main loop runs (this glue code is currently written in C, thus the need for the raw pointers).

### DIFF
--- a/wayland-client/src/display.rs
+++ b/wayland-client/src/display.rs
@@ -95,6 +95,12 @@ impl Display {
     }
 
     #[cfg(feature = "native_lib")]
+    /// Gets the raw wl_display pointer.
+    pub fn as_ptr(&self) -> *mut wl_display {
+        self.inner.display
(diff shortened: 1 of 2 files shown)


vberger:
The `Display` object already provides the `wl_display` pointer though the `c_ptr()` method of the `Proxy<WlDisplay>` it derefs to, I believe.

Regarding the event loop, I'm wary of exposing it without at least a clear warning: there should always be a single part of the program responsible of dispatching the event queue (it requires a `&mut` borrow after all). I guess in your case it would be the glib main loop that handles it?

Timidger:
Thanks for the feedback.

Looking into it more, I just need the `wl_display` for what I need to do, not the event queue. The `c_ptr` method already does that, so I'm going to go ahead and close this PR.

PR #180, opened 2018-07-07
Tidying in scanner

trimental:


vberger:
Thanks for this! :+1: 

I'm good with the first two commits. However, regarding the third one I have two concerns:

- I'm not sure I see the gain of moving from the `is_keyword` function to the `KEYWORDS` const array, could you explain why you suggest this change?
- In all cases, I'd rather not change the codegen part of the scanner right now, I've refactored it a lot in #170 and rebasing it would be rather painful...

trimental:
Hi, the hardcoding of keywords just seemed more inline with the concept of consts to me but it is a very small preference change and probably isn't worth the hassle. I removed that commit from the fork

PR #182, opened 2018-07-14
Proxy::equals -> PartialEq::eq, impl Eq for Proxy

### LINKED ISSUE #181: Implement PartialEq/Eq for Proxy<T>
Currently you can only campare `Proxy<T>` objects using the [equals](https://docs.rs/wayland-client/0.20.10/wayland_client/struct.Proxy.html#method.equals) method.

Instead the `PartialEq` and `Eq` traits should be used so that you can use the other methods that define on it. Case in point, I'm trying to use `contains` on a `Vec<Proxy<WlOutput>>` and it won't work.

I realize that this has a non-trivial cost/decision to make because it uses atomics internally, but that's already hidden from the user by this method.

Timidger:
This allows you to use functions which depend on the PartialEq and Eq trait being defined.

Fixes #181

### DIFF
--- a/wayland-client/src/proxy.rs
+++ b/wayland-client/src/proxy.rs
@@ -50,20 +50,8 @@ unsafe impl<I: Interface> Send for Proxy<I> {}
 unsafe impl<I: Interface> Sync for Proxy<I> {}
 
 impl <I: Interface> PartialEq for Proxy<I> {
-    /// Check if the other proxy refers to the same underlying wayland object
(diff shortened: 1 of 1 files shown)


vberger:
I'm ok with adding this, but for very simple reasons (the rebase of #170 which I believe will be nearly finished), I'd rather have the `PartielEq` implementation just be a proxy for the `equals()` method for now, if that's okay.

PR #187, opened 2018-07-19
Fuzzing

### LINKED ISSUE #184: Fuzzing the protocol parser?
Now that there is a rust implementation of the protocol, we also have [a parser for it](https://github.com/Smithay/wayland-rs/blob/master/wayland-commons/src/wire.rs).

I'd be tempted to try to setup a way to fuzz it, but I'm not clear about how this should be done, given the parsing depends on the signature which is supposedly known beforehand and tracked by the protocol state.

The function of interest is `wayland_commons::wire::Message::from_raw`

newpavlov:
Fuzzing with `cargo fuzz`. Closes #184.

To launch fuzzer use `cargo fuzz run message_parser` (you also can use `-j` option to increase number of workers). Currently it always passes 4 `RawFd`s and 16 `Argument`s, but I think it should be sufficient for testing.

It immediately found panic at `split_at` due to the insufficient length check. I will leave fuzzer for several hours, but looks like no other panics should happen.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,25 @@
+
+[package]
+name = "wayland-test-fuzz"
+version = "0.0.1"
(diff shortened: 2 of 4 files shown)


vberger:
Wow, thanks!

Issue #196, opened 2018-08-09
Redesign the `user_data` mechanism


vberger:
A very useful functionality  for implementing elaborate servers and client is to associate data to a wayland object (and to clients for server-side).

Currently, this is done via the `set_user_data` and `get_user_data` mechanism, which are obviously terrible as they only deal with `*mut ()`. The goal would be to add a better, more typed interface.

A few questions/constraints remain:

- should it be possible to associate several values to a given object ? (in a typemap kind of way)
  - server-side will need it to be able to associate several values to clients
- the runtime cost should be minimal is no value is actually associated
- to be safe, associated values must be `Send`, as the handles to wayland objects are themselves `Send`, and access to data must be synchronized
  - likely we will need an `unsafe` escape-hatch for non-send data, in the same way that non-`Send` implementations are possible

Drakulix:
From my perspective a typemap is unnecessary, since you may store your own, if you really need one. I would rather keep the code as simple as possible.

vberger:
The need for associating multiple data with `Client`s server-side is real though, even the C libraries have a similar mechanism.

And for objects, I think the bare minimum would be to add typing to the user_data, in an `Any`-like fashion, and with proper synchronization.

vberger:
One option I see for the Proxy/Resource user data would be to have it be set at implementation time (which is guaranteed to occur exactly once before the object is usable). The user data type `UD` would have to be `Send + Sync`, and a `user_data()` method of the proxy/resource would give access to it as a `&UD`. The user can then put a mutex or whatever they want to get interior mutability.

The downside of that is that it completely crosses-out the possibility to put non-`Send` user data (unless creating a wrapper and `unsafe impl`-ing `Send` and `Sync`, which is horrible from an ergonomics point of view.

Or I could introduce a similar "nonsend" dance with the `QueueToken`/`LoopToken`, but to access the user data. Meaning if the user data you stored is not send, you must assert you are accessing it from the correct thread using the token...

vberger:
`Proxy` and `Resource` user data has been done in #198, remains the question of `Client` associated user data.

The problem with `Client` is the question of associating several values to a single `Client`. This will be needed for several modules of `smithay`, or generally anyone that wants to develop a modular compositor. So the question is do we handle this multi-values in wayland-server or let the user handle it themselves (and use a simple `UserData` like for `Resource`s) ? For the record the C wayland libs do provide a mechanism to associate several values to a single `wl_client`.

[...]

PR #198, opened 2018-08-15
Rework Proxy&Resource user_data mechanism

vberger:
cc #196 

This changes the user data mechanism to be the following:

- the user data of a wayland object is set once and for all at implementation time
- it can be accessed from any handle via a type `user_data<UD>(&self) -> Option<&UD>` method which is a thin wrapper around `Box<Any>::downcast_ref()`
- the user is responsible for integrating interior mutability to the type if they want

In particular, the previous behavior can be reached by using an `AtomicPtr` as the `UD` type, but it makes the default behavior much more safe.

What do you think of this, @Drakulix ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - [protocols] Remove support for wayland-wall, which has been discontinued.
 - [client] Add `Display::get_display_ptr()` to differentiate between the wrapper and the
   actual `wl_display`
+- [client] Rework user-data mechanism to introduce type-safety
 
 # 0.21.0-alpha1 - 2018-07-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - [client] Add `Display::get_display_ptr()` to differentiate between the wrapper and the
   actual `wl_display`
 - [client] Rework user-data mechanism to introduce type-safety
+- [server] Rework `Resource` user-data mechanism to introduce type-safety
 
 # 0.21.0-alpha1 - 2018-07-18
 
--- a/tests/attach_to_surface.rs
+++ b/tests/attach_to_surface.rs
@@ -118,11 +118,11 @@ fn attach_null() {
 
     let compositor = manager
         .instantiate_exact::<wayc::protocol::wl_compositor::WlCompositor, _>(1, |comp| {
-            comp.implement(|_, _| {})
+            comp.implement(|_, _| {}, ())
         })
         .unwrap();
     let surface = compositor
-        .create_surface(|surface| surface.implement(|_, _| {}))
+        .create_surface(|surface| surface.implement(|_, _| {}, ()))
         .unwrap();
     surface.attach(None, 0, 0);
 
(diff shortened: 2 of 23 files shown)


Drakulix:
Overall this is very good, but this basically requires using a `Mutex`/`RwLock` for anything Non-Send/Sync, right?

What about making `UD` not require `Sync+Send`, but for `implement` (not for `implement_non_send` though). Then make `get_user_data` take a `QueueToken` to verify, that it is accessed from the correct thread. For implementations initialized using `implement`, that satisfy the Send+Sync requirement the Token is not checked or might be `None`.

I am not entirely sure, but I think I had several use cases for Non-Send userdata in `fireplace` (as well for interior mutability, but this abstractions seems fine in that case).

At lease we should provide an unsafe function `get_userdata_unchecked` that lifts the Send+Sync requirement.

vberger:
I think it's not possible to make non-Send user data work without exposing an `unsafe` function.

The main issue is that, at least with the `native_lib` backend, it is not possible given a `QueueToken` to check if a `Proxy` is associated to this event queue. The issue is the same server-side, I have no reliable way to check if a `LoopToken` does actually match the `Resource`.

Though, given `implement_nonsend` is already unsafe, it's not a big deal to add an unsafe `user_data_nonsend` method.

vberger:
> Is this a limitation we should bring up on the wayland mailing list? This appears to be an issue even in C, when one would try to implement a multi-queue compositor.

There is no notion of event queues server-side, all objects are on the same event loop. But this can technically be fooled by providing a loop token from an other event loop (not hosting wayland objects).

So, this may arguably fall under the "you're doing C you're supposed to know what you are doing" side of things.. ?

Drakulix:
> So, this may arguably fall under the "you're doing C you're supposed to know what you are doing" side of things.. ?

Yeah, probably..

Is there a way to do better for your pure-rust-implementation? Maybe we should provide another safe `userdata_nonsend` variant, that takes a loop token for the rust-impl only?

vberger:
It might be possible to do better with it, but I'm reluctant to go this direction. Given how cargo features work, activating the `native_lib` feature should be purely additive in terms of functionalities, and should not break any code.

Drakulix:
> It might be possible to do better with it, but I'm reluctant to go this direction. Given how cargo features work, activating the native_lib feature should be purely additive in terms of functionalities, and should not break any code.

Then just add another userdata-function instead of replacing the unsafe variant?

vberger:
The issue is that this new method would have to become unavailable when the `native_lib` feature is activated, making code using it fail to compile. This is not the expected behavior of cargo features, as cargo is free to activate more features than you requested, to match other dependencies of the build process.

Drakulix:
> The issue is that this new method would have to become unavailable when the native_lib feature is activated, making code using it fail to compile.

I am taking about an additional function only if `native_lib` is activated:

```
    pub fn user_data<UD: Send + Sync + 'static>(&self) -> Option<&UD> {
        unsafe { self.inner.get_user_data() }
    }
    
    pub unsafe fn user_data_nonsend<UD: 'static>(&self) -> Option<&UD> {
        self.inner.get_user_data()
    }

    #[cfg(feature = "native_lib")]
    pub fn user_data_token<UD: 'static>(&self, token: &LoopToken) -> Option<&UD> {
        if token.does_not_match() { panic!(); }
        unsafe { self.inner.get_user_data() }
    }
```

I do not see how this might make code not compile if the native_lib feature is used.

vberger:
That's the opposite, the safe function can only be done if we do *not** use the `native_lib`. When `native_lib` is activated, we are using the system C libs, which do not allow to check if the queue/loop match.

Drakulix:
I somehow assumed the feature would enable the rust implementation. :man_facepalming: 
I get the decision, why you wanted to make the c-library the `feature` (to remove it from the default list in the future), but given our goal of an ergonomic rust api, I am not sure if it was the best approach.

vberger:
The reason is actually because of how cargo features work : enabling a feature cannot be a breaking change.

A client using the rust implementation can also work with the system lib, but a client requiring the system lib cannot use the rust impl (as it uses the APIs exposing C pointers), so using the system libs must be the optional feature.

vberger:
Other approach, I introduce a small `UserData` wrapper in wayland-commons. It can be created in two variants, threadsafe requiring the data to be `Send+Sync`, and non-threadsafe with no constraints.

If created in the threadsafe variant, its contents can be accessed from any thread, if not, the `ThreadId` is checked and access is only possible from the creation thread.

I think this makes a better API, what do you think @Drakulix ?

Drakulix:
> I think this makes a better API, what do you think @Drakulix ?

Sounds good to me!

PR #205, opened 2018-09-16
Use sccache to speed-up travis builds.

vberger:
Try out https://twitter.com/oetzn/status/1041099844884938752 to see if it helps.


vberger:
`cargo-update` does not compile on old stable rust, this is getting too convoluted for its own good.

PR #210, opened 2018-10-21
wayland-scanner: fix missing closing paren in Argument::Array

myfreeweb:


myfreeweb:
Added a test (to an existing request, hope that's fine)

PR #213, opened 2018-11-08
Use quote for code generation in wayland-scanner

jplatte:
Posting this now to have a travis run for the first converted module and to make it easier to review. The other modules should look similar after conversion.

### DIFF
--- a/tests/scanner.rs
+++ b/tests/scanner.rs
@@ -10,10 +10,10 @@ use wayland_scanner::Side;
 const PROTOCOL: &'static str = include_str!("./scanner_assets/protocol.xml");
 
 const C_INTERFACES_TARGET: &'static str = include_str!("./scanner_assets/c_interfaces.rs");
-
 const CLIENT_C_CODE_TARGET: &'static str = include_str!("./scanner_assets/client_c_code.rs");
-
 const SERVER_C_CODE_TARGET: &'static str = include_str!("./scanner_assets/server_c_code.rs");
+const CLIENT_RUST_CODE_TARGET: &'static str = include_str!("./scanner_assets/client_rust_code.rs");
+const SERVER_RUST_CODE_TARGET: &'static str = include_str!("./scanner_assets/server_rust_code.rs");
 
 fn print_diff(diffs: &[Difference]) {
     println!("Partial diffs found:");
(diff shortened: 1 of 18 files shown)


jplatte:
Open question: Should the formatting of `tests/scanner_assets/*` be reverted to stay readable, and the tests for wayland-scanner be updated to format the `quote!` output before diff'ing with the assets (which would mean they would have to be skipped on older rustc versions, or possible anything that's not nightly) **or** should they remain in the hard-to-read `quote!` output style, with `rustfmt` only being used when available to improve the diff output in case of test failure?

jplatte:
For now, I've removed the copyright module doc comment since it is not actually working as expected. I'm not sure if this can be fixed easily.

vberger:
Regarding the formatting of the test files, I think it's better to keep them in a pretty formatted state. Otherwise, it makes it almost impossible to review any changes made to them in the future.

vberger:
Given I have merged your other PRs, could you rebase this?

tests/scanner.rs
```diff
@@ -75,59 +81,76 @@ fn flatten_diffs(diffs: &[Difference]) -> Vec<Difference> {
...
+        })
+        .collect()
+}
+
+fn run_codegen_test(generated_file_path: &Path, expected_output: &str) {
+    match Command::new("rustfmt")
+        .arg("--config-path")
+        .arg(env!("CARGO_MANIFEST_DIR"))
+        .arg(generated_file_path)
+        .status()
```
vberger: To better handle rustfmt changes, maybe we could:

- in the scanner tests, run `rustfmt` on both the generated file and the target file
- run `rustfmt --chek` on the target files in the `FMT` travis build

jplatte: Why both? If we `rustfmt --check` the test assets in CI, we'll have to update them anyway if rustfmt changes, also running it on them without `--check` as part of the test code seems like it would have no additional benefit while making the code (slighly) longer.

vberger: I agree it's not strictly needed. My intent was to ensure that only the rustfmt job would break if a rustfmt update changed the formatting.

jplatte:
[Travis failure](https://travis-ci.org/Smithay/wayland-rs/jobs/476597701) seems unrelated. Not sure what's going on there.

vberger:
Did you fully rebase on master ? There was a submodule update since you first opened your PR and it looks like travis is not fetching the proper commit of the submodules

jplatte:
Ah, I see. Yeah, seems like I accidentally reverted that in the rebase. Will get rid of that change.

vberger:
Thanks!

PR #216, opened 2018-12-18
Update formatting by re-running latest stable rustfmt

jplatte:
I noticed #213 recently failed because of formatting, and thought this would be good to do in a separate PR.

PR #217, opened 2018-12-20
Fix CI by re-running `cargo fmt` on stable, pinning cc version for rustc 1.21.0

jplatte:


jplatte:
#216 is enough after all since [cc 1.0.28](https://github.com/alexcrichton/cc-rs/compare/1.0.27...1.0.28) was just released with a much wider range of supported rustc versions.

PR #219, opened 2018-12-27
New protocols

dvc94ch:


vberger:
All is good, please just add a changelog entry for that :+1:

dvc94ch:
So we have the official wayland-protocols and the wlroots-protocols. Is there somewhere we can put qt/gtk/kde protocols that haven't been standardized yet or the xcursor-configuration protocol that is proposed for standardization in wayland-protocols?

In particular I'm interested in adding the xcursor-configuration protocol and the kwayland appmenu protocol.

vberger:
> Is there somewhere we can put qt/gtk/kde protocols that haven't been standardized yet

Are the KDE / Gnome protocols you speak of exposed somewhere publicly in a repo, like the wlr protocols? If so, we can integrate them the same way wlr protocols are integrated.

If not, I'd like to first be sure these protocols are indeed considered as public interfaces by their respective projects. If they are not, we would be subject to unexpected changes and incompatibilities between different versions of KDE / Gnome making them very painful to support.

> or the xcursor-configuration protocol that is proposed for standardization in wayland-protocols?

Can't that simply wait for the protocol to be standardized? Tbh I didn't really follow its recent evolution.

dvc94ch:
Ok, we can wait on xcursor-configuration, and the appmenu protocol I can support in a different crate, requires some dbus too.

vberger:
Alright, thanks!

PR #220, opened 2019-01-05
Implement Error for client errors

pyfisch:
This makes it easier to use `?`
as errors are automatically converted.

### DIFF
--- a/wayland-client/src/display.rs
+++ b/wayland-client/src/display.rs
@@ -41,9 +41,9 @@ impl fmt::Display for ConnectError {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
         use self::ConnectError::*;
         f.write_str(match self {
-            NoWaylandLib => "the `libwayland-client.so` library can not be found",
+            NoWaylandLib => "the `libwayland-client.so` library could not be found",
             XdgRuntimeDirNotSet => "the `XDG_RUNTIME_DIR` variable is not set while it should be",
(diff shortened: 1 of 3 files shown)


pyfisch:
The `enum GlobalError` should be changed as well.

wayland-client/src/display.rs
```diff
@@ -35,6 +37,21 @@ pub enum ConnectError {
     InvalidFd,
 }
 
+impl fmt::Display for ConnectError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        use self::ConnectError::*;
+        f.write_str(match self {
+            NoWaylandLib => "the `libwayland-client.so` library can not be found",
+            XdgRuntimeDirNotSet => "the `XDG_RUNTIME_DIR` variable is not set while it should be",
+            NoCompositorListening => "the compositor is not listening",
```
vberger: I'd prefer something like "could not find a wayland compositor listening", it'd better reflect the way things are under the hood imo.

vberger:
> The enum `GlobalError` should be changed as well.

This can be done in the same PR if you want.

pyfisch:
Updated.

pyfisch:
Travis failed for Rust v1.21.0 because `description` is missing on the Error trait. In the more recent Rust versions `description` is soft deprecated and just prints "description() is deprecated; use Display". Should I add the same placeholder here or do you want a proper description?

PR #224, opened 2019-01-18
Update wlr-protocols

YaLTeR:
Includes the new `data-control-unstable-v1` protocol.

I hit an issue trying to use it though. I have the following code:
```rust
use std::sync::{Arc, Mutex};

use wayland_client::{
    protocol::{
        wl_registry::RequestsTrait as WlRegistryRequestsTrait, wl_seat::WlSeat,
    },
    Display, GlobalEvent, GlobalManager, Interface,
};
use wayland_protocols::wlr::unstable::data_control::v1::client::{
    zwlr_data_control_manager_v1::{
        RequestsTrait as ZwlrDataControlManagerV1RequestsTrait,
        ZwlrDataControlManagerV1,
    },
    *,
};

fn main() {
    let (display, mut queue) =
        Display::connect_to_env().expect("Error connecting to a display");
    println!("Connected!");

    let seats = Arc::new(Mutex::new(vec![]));

    let global_manager = {
        let seats = seats.clone();
        GlobalManager::new_with_cb(&display, move |event, proxy| {
            if let GlobalEvent::New {
                id,
                interface,
                version,
            } = event
            {
                if interface == WlSeat::NAME && version >= WlSeat::VERSION {
                    seats.lock().unwrap().push(
                        proxy
                            .bind::<WlSeat, _>(WlSeat::VERSION, id, |seat| {
                                seat.implement(|_, _| (), ())
                            })
                            .unwrap(),
                    );
                }
            }
        })
    };

    queue.sync_roundtrip().expect("Error doing a roundtrip");

    let data_control_manager = global_manager
        .instantiate_exact::<ZwlrDataControlManagerV1, _>(
            ZwlrDataControlManagerV1::VERSION,
            |new_proxy| new_proxy.implement(|_, _| (), ()),
        )
        .expect("Error getting zwlr_data_control_manager_v1");

    for seat in &*seats.lock().unwrap() {
        data_control_manager
            .get_data_device(seat, |data_device| {
                data_device.implement(
                    |event, _| {
                        if let zwlr_data_control_device_v1::Event::DataOffer { id } =
                            event
                        {
                            id.implement(
                                |event, _| {
                                    let zwlr_data_control_offer_v1::Event::Offer {
                                        mime_type,
                                    } = event;
                                    println!("{}", mime_type);
                                },
                                (),
                            );
                        }
                    },
                    (),
                )
            })
            .unwrap();
    }

    queue.sync_roundtrip().expect("Error doing a roundtrip");
}
```
[...]

PR #225, opened 2019-01-18
commons: fix server ids handling in map

vberger:
Client ids start at 1 and server ids start at SERVER_ID_LIMIT, not
SERVER_ID_LIMIT+1.

cc #224


vberger:
The coverage issue is linked to https://github.com/xd009642/tarpaulin/issues/190, I'll ignore it.

Issue #228, opened 2019-01-21
Fundamental API discussion, yet again


vberger:
So, apparently there is friction with the current API. This API is mostly the result of my use of the libs in smithay and winit, with some other input. I will not claim I have the *best* idea of how the API should be designed, and I'm willing to discuss further the API if people have better ideas. So I'll detail here the trade-offs as I perceive them, as a starting point of the discussion if anyone wants to continue on this (cc @Timidger @Drakulix @YaLTeR @dvc94ch I believe you may be interested in this).

## Lexicon

- a wayland object **interface** is how it is defined in the protocol file. It is a name (eg `wl_buffer`), and a list of messages description. Messages can be **requests** (sent by the client to the server) or **events** (sent by the server to the client), but apart from the name difference they are completely symmetric.
- What I call a wayland object **implementation** is the code that, given a received message, processes it (currently implementations are closures `FnMut(MsgEnum)-> ()`)

## Fundamental required properties

So, the main API question revolves around how messages are *received*. The wayland protocol associates each message to a wayland object, and we have these fundamental properties we want to keep:

- **1)** two wayland objects of the same interface may have different implementation
- **2)** even if two objects have the same implementation, they may use object-specific data
- **3)** it must be possible to share data between an implementation and the rest of the program

Currently, these properties are held like so: each wayland object is independently implemented by a closure, which takes care of **1** and **2** (object-specific data is the closure state). **3** is done by either having `Rc`/`Arc` in the closure state, or using the `user-data` API.

Additionally there are some other properties we need to keep in mind:

- The wayland server **must** process messages from a given client **sequentially** and **in order**.
- A protocol error is always fatal to the wayland connection.
- It should be reasonably easy for users to use protocol extensions, even if they are not part of `wayland-protocols`

## Empirical ergonomic considerations

I've seen and experimented with two main approaches for handling the message reception and processing. Iterator-like, and callback-like.

### Iterator-like API

Event queues are iterator-like constructs that yield `(Object, Message)` tuples, linking each message with a handle to the object it is associated with.

**Pros**:

- works well with rust in general, notably wrt to state sharing (no need for rc or locking)

**Cons**:

- managing object-specific data or implementation require looking up something stored in the object's `user-data` to identify it
- the iterator return type must be able to express all potential messages:
  - this is difficult to integrate with the plug-in way of handling wayland protocols
[...]

PR #230, opened 2019-01-22
Trait-based event handling

YaLTeR:
As discussed in #228 (PoC v2 + `implement_closure()` + `implement_dummy()`).

Implemented client- and server-side.

Additionally I changed the `dynamic_globals` example to use trait-based event handling for `wl_output` as I thought it was a good place where the closure implementation was rather complex with lots of enum cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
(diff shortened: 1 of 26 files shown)


vberger:
Well, this all looks pretty neat, thanks for taking the time to put that together!

Could you please just add a changelog entry describing the change?

PR #232, opened 2019-01-23
Refactor to remove RequestsTrait and much of Proxy<_> wrapping

YaLTeR:
As proposed in https://github.com/Smithay/wayland-rs/issues/228#issuecomment-456297566.

This refactor removes `RequestsTrait`s and moves their methods onto the interface objects themselves. The interface objects contain the corresponding proxies. The end results are:
- It is no longer needed to use `RequestsTrait`s of all objects that requests are sent from, thus lots of boilerplate and renaming (all `RequestsTrait`s clash with each other) is avoided.
- A lot of `Proxy<Interface>` types in the client code are replaced with simply `Interface` which makes the types less complex and easier to understand.

Update of examples and tests is separated into its own commit, check it to see the client code changes.

TODO: update top-level docs on `wayland-client` and the changelog.

### DIFF
--- a/wayland-client/src/lib.rs
+++ b/wayland-client/src/lib.rs
@@ -150,7 +150,8 @@ pub mod cursor;
 #[cfg(feature = "egl")]
 pub mod egl;
 
-pub use wayland_commons::{AnonymousObject, Interface, MessageGroup, NoMessage};
+pub use anonymous_object::AnonymousObject;
+pub use wayland_commons::{Interface, MessageGroup, NoMessage};
 
 // rust implementation
 #[cfg(not(feature = "native_lib"))]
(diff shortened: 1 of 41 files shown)


YaLTeR:
I guess I also need to add some identifiers into the list of keywords in scanner, like `as_proxy`, so that it doesn't clash with a potential request method name.

wayland-scanner/src/common_gen.rs
```diff
@@ -626,11 +626,50 @@ pub(crate) fn gen_interface(
...
     addon: Option<TokenStream>,
+    side: Side,
 ) -> TokenStream {
+    let object_type = Ident::new(
+        match side {
+            Side::Client => "Proxy",
+            Side::Server => "Resource",
+        },
+        Span::call_site(),
+    );
```
vberger: We already have functionality for that: the `object_name()` method of `Side` returns an `Ident` with the proper value.

wayland-scanner/src/common_gen.rs
```diff
@@ -626,11 +626,50 @@ pub(crate) fn gen_interface(
...
-        pub struct #name;
+        #[derive(Clone, Eq, PartialEq)]
+        pub struct #name(#object_type<#name>);
+
+        impl #name {
+            #as_object_doc_attr
+            pub fn #as_object_name(&self) -> &#object_type<Self> {
+                &self.0
+            }
+        }
```
vberger: Maybe to avoid name clashes and to be more generic overall, this could be an implementation of the [`AsRef`](https://doc.rust-lang.org/std/convert/trait.AsRef.html) trait ? So an `impl AsRef<Proxy<#name>> for #name { ... }`.

YaLTeR:
> I see this PR only does the change from `Proxy<I>` to I client-side, not server-side. I don't know if it is just "to be done later" or not, but I think I prefer that the switch is done server-side as well, so to avoid having a lot of special casing on side in the scanner.

I skipped the serverside part because there was no equivalent to `RequestsTrait` and after a quick look it seems to work slightly differently (no `Resource::child_placeholder()`, etc.). Could you provide serverside versions of stuff from `RequestsTrait`?

wayland-scanner/src/common_gen.rs
YaLTeR: Oh yeah, good idea.

vberger:
Ah right. Well server-side is much easier: the methods for sending objects never create any object themselves, and just treat `NewId` arguments as if they were `Object`. The compositor writer is responsible for creating the objects manually and passing them as argument to the method.

The event-specific methods server-side would thus really just be thin layers on top of `Resource::send()`.

wayland-client/src/lib.rs
```diff
@@ -3,76 +3,74 @@
...
+//! Wayland object. Multiple instances of a Rust object can exist referring to the same
 //! protocol object.
 //!
-//! Similarly, the lifetimes of the protocol objects and the `Proxy<I>` are
+//! Similarly, the lifetimes of the protocol objects and the Rust objects are
 //! not tightly tied. As protocol objects are created and destroyed by protocol
 //! messages, it can happen that an object gets destroyed while one or more
-//! `Proxy<I>` still refers to it. In such case, these proxies will be disabled
+//! Rust objects still refer to it. In such case, these Rust objects will be disabled
 //! and their `alive()` method will start to return `false`. Trying to send messages
```
vberger: The `alive()` method is on `Proxy<_>`, not the rust object itself.

wayland-server/src/lib.rs
```diff
@@ -3,67 +3,73 @@
...
-//! subsequently sent to them are ignored.
+//! main difference being that the underlying handles to objects are represented by the `Resource<I>`
+//! type, very similarly to proxies in `wayland-client`.
+//!
+//! These resources are used to send messages to the client (in the Wayland context,
+//! these are called "events"). You usually don't use them directly, and instead call
+//! methods on the Rust objects themselves, which invoke the appropriate `Resource` methods.
+//! It is also possible to directly use the `Resource::<I>::send(..)` method, but
+//! this should only be done carefully: using it improperly can mess the protocol
+//! state and cause protocol errors.
```
vberger: Server side, the methods are really just shorthands for the `send(..)` method, and provide no additional safety. The same care should be used for both.

wayland-server/src/lib.rs
```diff
@@ -3,67 +3,73 @@
...
+//! There is not a 1 to 1 mapping between Rust object instances and protocol
+//! objects. Rather, you can think of the Rust objects as `Rc`-like handles to a
+//! Wayland object. Multiple instances of a Rust object can exist referring to the same
+//! protocol object.
+//!
+//! Similarly, the lifetimes of the protocol objects and the Rust objects are
+//! not tightly tied. As protocol objects are created and destroyed by protocol
+//! messages, it can happen that an object gets destroyed while one or more
+//! Rust objects still refer to it. In such case, these Rust objects will be disabled
+//! and their `alive()` method will start to return `false`. Events that are subsequently
```
vberger: Similarly, this is a method of `Resource<I>`

PR #234, opened 2019-01-24
Introduce event iterators & sinks

vberger:
This is the last outstanding point of #228 , this allows downstream users to opt-in to an iterator-based event handling for some of their Wayland objects.

Also introduces an adapter in wayland-client that links an event iterator to an event queue, so that it will block on it if there are no pending events.

### DIFF
--- a/wayland-client/src/sinks.rs
+++ b/wayland-client/src/sinks.rs
@@ -1,0 +1,165 @@
+//! Message sinks
+//!
+//! The Wayland model naturally uses callbacks to handle the message you receive from the server.
+//! However in some contexts, an iterator-based interface may be more practical to use, this is
+//! what this module provides.
+//!
+//! The `message_iterator` function allows you to create a new message iterator. It is just a
+//! regular MPSC, however its sending end (the `Sink<T>`) can be directly used as an implementation
+//! for Wayland objects. This just requires that `T: From<(I::Event, I)>` (where `I` is the
+//! interface of the object you're trying to implement). The `event_enum!` macro is provided to
+//! easily generate an appropriate type joining events from different interfaces into a single
+//! iterator.
+//!
+//! The `blocking_message_iterator` function is very similar, except the created message iterator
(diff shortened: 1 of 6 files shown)


wayland-client/src/sinks.rs
```diff
@@ -0,0 +1,165 @@
+//! Message sinks
+//!
+//! The Wayland model naturally uses callbacks to handle the message you receive from the server.
+//! However in some contextes, an iterator-based interface may be more practical to use, this is
```
YaLTeR:
```suggestion
//! However in some contexts, an iterator-based interface may be more practical to use, this is
```

wayland-client/src/sinks.rs
```diff
@@ -0,0 +1,165 @@
...
+//! what this module provides.
+//!
+//! The `message_iterator` function allows you to create a new message iterator. It is just a
+//! regular MPSC, however its sending end (the `Sink<T>`) can be directly used as an implementation
+//! for Wayland objects. This just require that `T: From<(I::Event, I)>` (where `I` is the
```
YaLTeR:
```suggestion
//! for Wayland objects. This just requires that `T: From<(I::Event, I)>` (where `I` is the
```

wayland-commons/src/sinks.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn clone(&self) -> Sink<T> {
+        Sink {
+            queue: self.queue.clone(),
+        }
+    }
+}
+
+/// A message iterator
+///
+/// I yields the various message that have been pushed to it from its associated
```
YaLTeR:
```suggestion
/// It yields the various messages that have been pushed to it from its associated
```

dvc94ch:
Sorry for ghosting :) been busy with other things. Having some experience with tokio and async/await on it's way to stabilization, I think it's the way forward. I'm unlikely to have time to work on this at the moment :/

PR #236, opened 2019-01-25
commons: close fds we dup-ed after sending them

### LINKED ISSUE #235: Transferred FDs aren't properly closed
Consider receiving data over a [zwlr_data_control_offer](https://smithay.github.io/wayland-rs/wayland_protocols/wlr/unstable/data_control/v1/client/zwlr_data_control_offer_v1/struct.ZwlrDataControlOfferV1.html):
```rust
// Create a pipe for content transfer.
let (mut read, write) = pipe().expect("Error creating a pipe");

// Start the transfer.
offer.receive(mime_type, write.as_raw_fd());
queue.sync_roundtrip().expect("Error doing a roundtrip");
drop(write);

// Read the contents.
let mut contents = vec![];
read.read_to_end(&mut contents)
    .expect("Error reading clipboard contents");

stdout().write_all(&contents)
        .expect("Error writing contents to stdout");
```
Using `native_lib` this works, using the Rust implementation this blocks on read. I highly suspect this is due to `wayland-commons` duping FDs before writing and subsequently never closing them over here: https://github.com/Smithay/wayland-rs/blob/d508d0e7988fbc0f3a859e927cf749747c6cf9e1/wayland-commons/src/wire.rs#L192-L224

Changing the last line to `drop(pending_fds)` isn't valid either because we can't close the FDs before the server received them.

vberger:
Fixes #235 

@YaLTeR can you confirm this fixes the issue ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@
   design change above.
 - [scanner] Fixed a number of cases where invalid Rust code would be generated if variable
   or method names in the protocol were Rust keywords.
+- [commons] Properly close FDs we dup-ed after sending them to the server, to avoid leaking
+  open FDs.
 
 ## 0.21.11 -- 2019-01-19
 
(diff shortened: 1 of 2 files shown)


YaLTeR:
Confirming it fixes the issue.

PR #241, opened 2019-01-29
GlobalManager updates

YaLTeR:
- drop the `Send` bounds,
- fix the warning in `global_filter!()`.


YaLTeR:
Good point. Changed it back.

vberger:
Looks good :+1:

Could you just move the note about making the display as a proxy in the documentation of the display manager rather than the changelog? I think this is where it'd be most easily found, and I'd rather not have the changelog entries be that long.

vberger:
Perfect, thanks!

PR #242, opened 2019-01-31
Future proof protocol updates

vberger:
- Future-proof protocol enums by adding a `#[doc(hidden)] __nonexhaustive` variant. Should be migrated to https://github.com/rust-lang/rust/issues/44109 once it is stabilized. This makes updating the protocols a non-breaking change.
- Remove `GlobalManager::instantiate_auto` which was way too error prone, and replace it with `GlobalManager::instanciate_range`. Also improve the docs regarding to interface versioning.
- Actually check that the versions of messages sent and received over the wire are correct wrt to their supported version, and trigger protocol errors if they aren't (as the C lib does).

### DIFF
--- a/wayland-client/src/globals.rs
+++ b/wayland-client/src/globals.rs
@@ -90,6 +90,7 @@ impl GlobalManager {
                             wl_registry::Event::GlobalRemove { name } => {
                                 inner.list.retain(|&(n, _, _)| n != name);
                             }
+                            _ => {}
                         }
                     },
                     (),
@@ -161,7 +162,8 @@ impl GlobalManager {
                                             name
                                         );
                                     }
-                            }
+                            },
+                            _ => {}
                         }
                     },
                     (),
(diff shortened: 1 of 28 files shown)


wayland-client/src/globals.rs
```diff
@@ -175,43 +177,66 @@ impl GlobalManager {
...
+                }
             }
         }
         Err(GlobalError::Missing)
     }
 
-    /// Instantiate a global with a specific version
+    /// Instantiate a global from a version range
+    ///
+    /// If you want to support several versions for a particular global, this method allows you to
```
YaLTeR:
```suggestion
    /// If you want to support several versions of a particular global, this method allows you to
```

wayland-client/src/globals.rs
```diff
@@ -175,43 +177,66 @@ impl GlobalManager {
...
+    /// specify a range of versions that you accept. It'll bind the highest possible version that
+    /// is between `min_version` and `max_version` inclusive, and return an error if the highest
+    /// version supported by the compositor is lower than `min_version`.
```
YaLTeR: Add the note about not using `I::VERSION` here too.

wayland-client/src/globals.rs
```diff
@@ -175,43 +177,66 @@ impl GlobalManager {
...
+    /// as such unless you specifically want to support several versions of a protocol, it is
+    /// recommended to use this method with an hardcoded value for the version (the one you'll
+    /// use a as reference for your implementation). Notably you should *not* use `I::VERSION`
+    /// as a version, as this value can change when the protocol files are updated.
     ///
     /// This method is only appropriate for globals that are expected to
     /// not exist with multiplicity (such as `wl_compositor` or `wl_shm`),
-    /// as it will only bind a single one.
-    pub fn instantiate_auto<I: Interface + From<Proxy<I>>, F>(&self, implementor: F) -> Result<I, GlobalError>
+    /// as it will always the first one that was advertized.
```
YaLTeR: Accidentally a word

PR #262, opened 2019-05-14
Improve consistency in docs, fix typos

jplatte:
I still had this lying around. I wanted to go through all of the modules, but I think this is already useful as-is.


vberger:
Oh, nice, thanks a lot!

PR #263, opened 2019-05-16
Inform cargo of build script file / env var dependencies

### LINKED ISSUE #255: build script should notify cargo of their dependency on protocol files
Quoting @jplatte:

> You want to `println!("cargo:rerun-if-changed={}", protocol_file);` (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script)

So that the build script is re-run if the protocol file changes

jplatte:
fixes #255

### DIFF
--- a/wayland-client/build.rs
+++ b/wayland-client/build.rs
@@ -10,5 +10,6 @@ fn main() {
     let out_dir_str = var("OUT_DIR").unwrap();
     let out_dir = Path::new(&out_dir_str);
 
+    println!("cargo:rerun-if-changed={}", protocol_file);
     generate_code(protocol_file, out_dir.join("wayland_api.rs"), Side::Client);
 }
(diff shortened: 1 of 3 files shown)


wayland-client/build.rs
```diff
@@ -10,5 +10,6 @@ fn main() {
     let out_dir_str = var("OUT_DIR").unwrap();
     let out_dir = Path::new(&out_dir_str);
 
+    println!("cargo:rerun-if-changed={}", protocol_file);
```
vberger: You need to call `.display()` on the paths to get an object that can actually be fed to `println!()`.

vberger:
Apart from the `Display` issue, lgtm. Thanks! :+1:

wayland-client/build.rs
jplatte: This is not the part of the code where `protocol_file` is a `&Path`, but you're right of course. I apparently didn't test locally :man_facepalming:

vberger:
Thanks!

PR #267, opened 2019-06-15
Set Proxy::implement_dummy threadsafe

timothee-haudebourg:
Since `|_, _| ()` and `()` are both `Send`, I see no reasons why `Proxy::implement_dummy` shouldn't use `implement_closure_threadsafe` instead of `implement_closure`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 0.23.6 -- 2019-06-16
+
+- [client/server] Make `NewProxy/NewResource::implement_dummy()` threadsafe.
+
 ## 0.23.5 -- 2019-06-13
 
 - Update `nix` dependency to 0.14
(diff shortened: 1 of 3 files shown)


vberger:
Thanks for pointing this oversight. Could you do the same change in `wayland-server`, to keep the two crates in sync, and add a changelog entry?

timothee-haudebourg:
Here it is. I hope I've respected the changelog conventions.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 0.23.6 -- 2019-06-16
```
vberger: Please leave this in a "Unreleased" section, version 0.23.6 has not yet been released. ;)

timothee-haudebourg: Oops, I don't know what I was thinking. I hope I got it right this time.

vberger:
Thanks!

Issue #189, opened 2018-07-19
Expose `libwayland`-compatible C ABI


vberger:
The goal would be to be able to inject our own symbols to replace the system libwayland.

Typically, this might be doable with something like that (presenting client-side only, but its symmetric for server-side):

- a low-level crate creating a `libwayland-client.so` dynamic library that exposes both
  - a low-level rust ABI (similar to the api of the current `imp` module)
  - a C ABI compatible with freedesktop's `libwayland-client.so`
- `wayland-client` dynamically links to this previous crate using its rust ABI to expose the higher level API we currently know

There are still uncertainties about how this may be done / if it may work:

- will this convince the system linker that `libwayland-client.so` is already loaded, and that it does not need to load the system one?
- how does this interact with cargo, and the distribution of crates using `wayland-client`? There is now a dynamic library required for running them.

A possible (?) alternative would be to find a way to embed the C ABI symbols into the final static executable produced by rustc and somehow convince the linker that it does not need to load the system lib and rather use them. No idea if that's actually possible though.

PR #268, opened 2019-06-23
Use smallvec for message arguments

### LINKED ISSUE #249: Message type optimizations
I think `wayland_commons::wire::Message` is a bit sub-optimal right now.

Currently `Vec` is allocated for every message, while judging by [`wayland.xml`](https://github.com/Smithay/wayland-rs/blob/master/wayland-server/wayland.xml) max number of arguments is 8. To remove unnecessary allocations we can use `SmallVec` for `args`, though we will have to decide the array size for `SmallVec`, e.g. 3, 4, 6, or 8.

`Argument` always heap allocates for `Array` and `String` types, not only it causes needless allocations and data copies, but also makes `Argument` bigger (right now it uses 32 bytes). Instead of copying data we could point to the origin buffer instead, i.e. we will add a lifetime `Argument<'a>` bound to the origin buffer (and consequently to `Message` as well). The simplest option will be to use `&'a [u8]` and `&'a CStr`, but it will bring `Argument` size to only 24 bytes. Instead with a bit of unsafe code we can store `*const u8` and phantom data (to track lifetime) inside two opaque types which will dereference to `&'a [u8]` and `&'a CStr` respectively. It will bring `Argument` size to 16 bytes on 64-bit machines. I think adding lifetime to `Message` shouldn't be a big problem, as IIUC they are almost immediately consumed inside event loop.

newpavlov:
Closes #249 

Unfortunately it's a backwards incompatible change, at lease for `wayland-commons` and probably for other crates which use it in public API.

Maybe use this chance to update to 2018 edition?

### DIFF
--- a/wayland-client/Cargo.toml
+++ b/wayland-client/Cargo.toml
@@ -20,6 +20,7 @@ nix = "0.14.1"
 downcast-rs = "1.0"
 bitflags = "1.0"
 libc = "0.2"
+smallvec = "0.6"
 calloop = { version = ">=0.3.1, <0.5", optional = true }
 mio = { version = "0.6.0", optional = true }
 
--- a/wayland-client/src/lib.rs
+++ b/wayland-client/src/lib.rs
@@ -120,6 +120,8 @@ extern crate bitflags;
 extern crate downcast_rs as downcast;
 extern crate libc;
 extern crate nix;
+#[macro_use]
+extern crate smallvec;
 
 #[cfg(feature = "eventloop")]
 extern crate calloop;
--- a/wayland-commons/Cargo.toml
+++ b/wayland-commons/Cargo.toml
@@ -15,4 +15,4 @@ travis-ci = { repository = "Smithay/wayland-rs" }
 [dependencies]
 wayland-sys = { version = "0.23.5", path = "../wayland-sys" }
 nix = "0.14.1"
-
+smallvec = "0.6"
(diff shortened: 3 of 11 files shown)


newpavlov:
I am not sure why Rust 1.21 tests fail, `lazyvec` should have MSRV equal to 1.20.

vberger:
On principle I approve this PR :+1: , we just need to figure out why rust 1.21 fails. Looks like the generated scanner code has an issue somehow, unfortunately travis truncated the log, so this will need to be run locally.

Given this is going to be a breaking change, and with the possibility of going to rust 2018, this will likely happen, but I need to decide a few things, depending on if / how I want to address #189. So I'll likely let this PR sit a few days while I decide the path I want to take.

newpavlov:
For now I've bumped MSRV to 1.24, which is required by `lazycell v1.2`, and looks like it fixes the problem.

vberger:
Okay so, it has been quite more than a few days, but now I have a clearer idea.

#271 is nearly finished and will be some enormous breaking change, so this can be integrated with it.

Do you want to rebase your PR once I've landed #271, or shall I manually integrate your changes #271 ?

vberger:
Thinking back about this, I have one slight source of uneasiness: this would make smallvec part of the public API of `wayland-commons`, which I'm reluctant to do given it does not look like `smallvec` will go 1.0 soon.

PR #272, opened 2019-09-08
Format code using 'cargo fmt'

Atul9:


vberger:
Hi, thanks for the PR! It is true that the CI is currently failing due to fmt issues, probably and update from rustfmt that slightly changed the default options.

However, there is an other big PR (#271) which I plan to merge soon that is a big refactor of the library and its API. Merging your PR now would make this other PR difficult to rebase/merge, so I'll have to decline. Thanks for your interest though!

PR #274, opened 2019-09-12
Reinstall tarpauling for rust coverage

vberger:
A small try to see if it works better now.


vberger:
Lets trys again, as a new tarpaulin has been released potentially fixing the issue. :)

PR #280, opened 2019-11-15
Fixed copy-paste error in example

simvux:
Doesn't appear to actually make a difference but for the sake of correctness


vberger:
Thanks! This was indeed incorrect, even though it would only actually behave incorrectly if the compositor sends a second `capabilities` event, which is relatively rare.

PR #296, opened 2020-03-13
Resolve clippy warnings

namachan10777:
* Abolish deprecated method `description` of `::std::error::Error.`
 * Factorize complex type into `type` definitions in `wayland-protocols/build.rs`


vberger:
Thanks!

PR #303, opened 2020-04-17
[WIP] Make the `cursor` feature full rust

esposm03:
~~This PR inserts a Rust implementation into the wayland-cursor crate, but allowing to fallback to the system wayland-cursor.so, for interoperability with other libraries.
The rust implementation is backwards compatible for the most part (except for some lifetimes that aren't needed anymore).~~

This PR reimplements in Rust the `wayland-cursor` crate. The API is changed to be hopefully more Rust-y, since we don't have the c library to follow anymore.

### DIFF
--- a/wayland-cursor/src/lib.rs
+++ b/wayland-cursor/src/lib.rs
@@ -12,6 +12,32 @@
 //! means of querying which frame of the animation should be displayed at
 //! what time, as well as handles to the buffers containing these frames, to
 //! attach them to a wayland surface.
+//!
+//! # Example
+//!
+//! ```ignore
+//! use wayland_cursor::CursorTheme;
+//! # use std::thread::sleep;
+//! # use std::time::{Instant, Duration};
+//!
+//! let cursor_theme = CursorTheme::load(32, wl_shm);
+//! let cursor = cursor_theme.get_cursor("wait").expect("Cursor not provided by theme");
+//!
+//! let start_time = Instant::now();
+//! loop {
+//!     // Obtain which frame we should show, and for how long.
+//!     let millis = start_time.elapsed().as_millis();
+//!     let fr_info = cursor.frame_and_duration(millis as u32);
+//!
+//!     // Here, we obtain the right cursor frame...
+//!     let buffer = cursor[fr_info.frame_index];
+//!     // and attach it to a wl_surface.
+//!     cursor_surface.attach(Some(&buffer), 0, 0);
+//!     cursor_surface.commit();
+//!
+//!     sleep(fr_info.frame_duration);
+//! }
+//! ```
 
 use std::{
     env,
@@ -143,6 +169,7 @@ impl CursorTheme {
 pub struct Cursor {
     name: String,
     images: Vec<CursorImageBuffer>,
+    total_duration: u32,
 }
 
 impl Cursor {
@@ -152,16 +179,39 @@ impl Cursor {
     /// This will also grow `theme.pool` if necessary.
     fn new(name: &str, theme: &mut CursorTheme, images: &[xcur::parser::Image], size: u32) -> Self {
         let mut buffers = Vec::with_capacity(images.len());
+        let size = Cursor::nearest_size(size, images);
         let iter = images.iter().filter(|el| el.width == size && el.height == size);
 
         for img in iter {
             buffers.push(CursorImageBuffer::new(theme, img));
         }
 
+        let total_duration = buffers.iter().map(|el| el.delay).sum();
+
         Cursor {
+            total_duration,
             name: String::from(name),
             images: buffers,
         }
+    }
+
+    fn nearest_size(size: u32, images: &[xcur::parser::Image]) -> u32 {
+        let size = size as i32;
+        let mut all_sizes = Vec::new();
+
+        for img in images {
+            if !all_sizes.contains(&(img.width as i32)) {
+                all_sizes.push(img.width as i32);
+            }
+        }
+
+        let mut min = 0;
+        for (i, width) in all_sizes.iter().enumerate() {
+            if (width - size).abs() < (all_sizes[min] - size).abs() {
+                min = i;
+            }
+        }
+        all_sizes[min] as u32
     }
 
     /// Given a time, calculate which frame to show, and how much time remains until the next frame.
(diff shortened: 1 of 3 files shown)


esposm03:
So, I've marked this PR as WIP because there are still some points missing or needing discussion:
- Do we need to use `mmap` or `shmget` instead of a simple `tempfile::tempfile()`? (My understanding is that an opened file is an `mmap` pool anyways, so it shouldn't be needed, but that doesn't seem to be the case in other crates of this project).
- Better testing is needed. Basic loading of a single cursor into a buffer is working, but I didn't look at other things.

vberger:
Oh wow, this is neat ! :smile: 

I haven't yet looked at the code of your PR itself, but to answer your few points:

> This PR implements inserts a Rust implementation into the wayland-cursor crate, but allowing to fallback to the system wayland-cursor.so, for interoperability with other libraries.

Do you have any specific thing in mind about this? Because I've never hit a situation where interop wrt libwayland-cursor.so was needed. So if it is not needed, I'd say let's just drop it.

> Do we need to use `mmap` or `shmget` instead of a simple `tempfile::tempfile()`?

I think a good approach for creating the SHM file would be to follow SCTK approach, to avoid any interaction with the filesystem:

https://github.com/Smithay/client-toolkit/blob/c9b7cfdffdac3c8f724290585dbeabdbbe1f6f4b/src/shm/mempool.rs#L249

vberger:
Also, there are already breaking changes pending for the next release, so we don't need to worry too much about backward-compatibility. I'd rather focus on making a good API tbh.

esposm03:
Okay, I'll be looking into this tomorrow!

vberger:
On thing I just realized though. In the wayland-cursor API, the initial theme loading `load_theme` takes a `size: u32` argument. This argument represents the dimensions at which the cursor images should be loaded.

I don't know much about XCursor files, nor how they manage that (are they vector graphics?), but this par of the API is important and necessary for correct handling of cursor images to correctly behave in an HiDPI context, and to respect the `XCURSOR_SIZE` environment variable that is being standardized in Wayland environment.

However, your code does not appear to handle that, and actually `xcur` does not appear to support setting a size at all. Do you have a plan regarding this question? That is going to be a blocking point for merging your PR.

esposm03:
Yes, I had just realized that XCursor files can have multiple sizes, not a single one, and was working on it. Regarding the env variable, it's actually not cited in the xcursor man page, where I was reading from, but I checked the ArchWiki, and it is there. So, thanks for notifying me 😅

>  actually xcur does not appear to support setting a size at all

Yes, but I can get all the sizes and filter them manually.

vberger:
Ah ok. :+1: 

Regarding the env variables, currently wayland-cursor does not follow them at all, and it's up to the downstream users to read the env variables and call `load_theme` with the appropriate values.

Now, probably we could add in wayland-cursor a `load_theme_from_env` function that'd do just that, in addition to the current `load_theme`.

esposm03:
Hm, I believe maybe we should make that the default behavior.

vberger:
I don't get your point. There would just be two functions for loading a theme: `load_theme_from_env` that just loads the theme specified by the env (with fallback to `"default"`), and `load_theme` (or `load_theme_from_spec` if we want an explicit name) that load it from the values provided by the user.

How would one be more "default" than the other? We still need both, apss should be able to disregard the environment variables if they really need to.

esposm03:
Sorry for not being clear, I was just talking about making `load_theme` require a name, or load from env variable if that is found, and having something like `load_theme_no_env` to ignore the env variable.

vberger:
Ah, so have the function loading from env take a fallback name in case the env is not set?

Though, in that case, if an app wants to load cursor theme from the environment, why would it need to set the fallback to something else than `"default"`?

esposm03:
> Though, in that case, if an app wants to load cursor theme from the environment, why would it need to set the fallback to something else than "default"?

Sorry for now answering sooner, but I have added three methods on `CursorTheme`:
- `load`, which uses the environment variable or "default" as the name,
- `load_or`, which uses the environment variable or the provided name,
- `load_from_name`, which only uses provided values.

I believe these should be powerful enough, let me know what you think about this.

Also, I'm marking this as non-draft, as now I think the work is complete.

wayland-cursor/src/lib.rs
```diff
@@ -216,14 +198,161 @@ impl<'a> Cursor<'a> {
...
+        for pixel in pixels {
+            buf.push((pixel >> 24) as u8);
+            buf.push((pixel >> 16) as u8);
+            buf.push((pixel >> 8) as u8);
+            buf.push(*pixel as u8);
+        }
+
+        buf
+    }
+}
```
vberger, line 258: We can use [`byteorder`](https:://docs.rs/byteorder) to handle that. `wayland-client` already depends on it anyway.

wayland-cursor/src/lib.rs
```diff
@@ -216,14 +198,161 @@ impl<'a> Cursor<'a> {
...
 }
+
+/// Which frame to show, and for how long.
+///
+/// This struct is output by `Cursor::frame_and_duration`
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct FrameAndDuration {
+    /// The index of the frame which should be shown.
+    pub frame_index: usize,
+    /// The duration that the frame should be shown for.
```
vberger, line 274:
```suggestion
    /// The duration that the frame should be shown for (in milliseconds).
```

wayland-cursor/src/lib.rs
```diff
@@ -216,14 +198,161 @@ impl<'a> Cursor<'a> {
...
+            Err(nix::Error::Sys(Errno::ENOSYS)) => break,
+            Err(nix::Error::Sys(errno)) => return Err(io::Error::from(errno)),
+            Err(err) => unreachable!(err),
+        }
+    }
+
+    // Fallback to using shm_open
+    let sys_time = SystemTime::now();
+    let mut mem_file_handle = format!(
+        "/smithay-client-toolkit-{}",
```
vberger, line 312: Lets change that to `/wayland-cursor-rs-{}`, to keep our scoping correct.

wayland-cursor/src/lib.rs
```diff
@@ -216,14 +198,161 @@ impl<'a> Cursor<'a> {
...
+                    Ok(_) => return Err(io::Error::from(errno)),
+                    Err(nix::Error::Sys(errno)) => return Err(io::Error::from(errno)),
+                    Err(err) => panic!(err),
+                },
+                Err(err) => panic!(err),
+            },
+            Err(nix::Error::Sys(Errno::EEXIST)) => {
+                // If a file with that handle exists then change the handle
+                mem_file_handle = format!(
+                    "/smithay-client-toolkit-{}",
```
vberger, line 333: Same here.

wayland-cursor/src/lib.rs
m-hugo: > We can use [`byteorder`](https:://docs.rs/byteorder) to handle that. `wayland-client` already depends on it anyway.

How did github go from https:://docs.rs/byteorder to
https://github.com/Smithay/wayland-rs/pull/://docs.rs/byteorder to
https://github.com/Smithay/wayland-rs/compare/:/docs.rs/byteorder?expand=1
even now it continues to surprise me every day

esposm03:
Okay, everything should be okay now!

esposm03:
Oh and also, should I bump the version? Because if this gets published as-is, I believe there may be problems with client-toolkit depending on the old version.

vberger:
> Actually, should I bump the version? Because if this gets published as-is, I believe there may be problems with client-toolkit depending on the old version.

No need, I'll bump the version when I release a new version to crates.io.

esposm03:
Okay👍

vberger:
Alright, all is good, this is just missing a changelog entry before I merge.

esposm03:
Oh, right🤦.

esposm03:
Done!

vberger:
Thanks!

Issue #223, opened 2019-01-16
Properly implement 'prepare_read' in full-rust wayland-client


vberger:
Currently, `EventQueue::prepare_read()` is partially mocked in the pure-rust backend of wayland-client. The proper behavior, matching the C library is that:

- `prepare_read` increases an internal counter of threads getting ready to read
- when `read_events()` is called, the threads are blocked on a fence until all threads either are currently blocked or have cancelled their read
- at that point, one threads is unlocked and proceeds to the actual read, and then all other threads are released with "nothing to read yet" return value

Right now, `read_events` just acquires an exclusive lock on the socket, and reads from it. meaning if several threads try to read at the same time, 1 will read, and then the other will also try to read (the syscall will be sent and processed) to discover that there is nothing to read.

This is not a safety issue, but a performance one. Though it should likely only be an issue in applications trying to read events from different threads concurrently and are quite performance sensitive wrt to that.

PR #312, opened 2020-05-05
dispatch_pending is &self for both rust/C

Matthias-Fauconneau:
It is useful to allow dispatch_pending(&self, ...).
For example while another reference is kept for prepare_read/read_events polling.

### DIFF
--- a/wayland-client/src/event_queue.rs
+++ b/wayland-client/src/event_queue.rs
@@ -154,7 +154,7 @@ impl EventQueue {
     ///
     /// If an error is returned, your connection with the wayland compositor is probably lost.
     /// You may want to check `Display::protocol_error()` to see if it was caused by a protocol error.
-    pub fn dispatch_pending<T: std::any::Any, F>(&mut self, data: &mut T, fallback: F) -> io::Result<u32>
+    pub fn dispatch_pending<T: std::any::Any, F>(&self, data: &mut T, fallback: F) -> io::Result<u32>
     where
         F: FnMut(RawEvent, Main<AnonymousObject>, DispatchData<'_>),
     {


vberger:
Semantically it is not really. Allowing it to be `&self` would allow to call `dispatch_pending` from within a `dispatch_pending` itself, which would result in a deadlock.

What are you trying to achieve with that?

Matthias-Fauconneau:
To me since the dispatch receiver is symmetric to the blocking read_events sender, then it would be consistent for it to be blocking as well.
I stumbled on this because I have a ref I'm moving in a SelectAll and another to dispatch
https://github.com/Matthias-Fauconneau/iced/blob/sctkshm/sctk/src/async_sctk/mod.rs#L131-L155
i.e My use explicitly maps the Sender/Receiver situation of read_events/dispatch.
Actually, since multiple references to a shared buffer is not self-contained, the logic I'd really want is to just have a read_message which yield an opaque Message which I can forward through the stream to a dispatch_message.

vberger:
> To me since the dispatch receiver is symmetric to the blocking read_events sender, then it would be consistent for it to be blocking as well.

I'm sorry, but I don't get this. Neither `read_events` nor `dispatch_pending` are blocking. The reason the `dispatch*` methods take a `&mut` reference is to properly reflect what is allowed or not. Allowing `&self` access would allow these functions to be called recursively, which would lead to a deadlock or a crash, depending on the context. If you need shared access to the `EventQueue`, the proper way would IMO to use a `RefCell<EventQueue>`.

> the logic I'd really want is to just have a read_message which yield an opaque Message which I can forward through the stream to a dispatch_message.

I'd like that too, but that is not possible due to the C lib API not exposing a way to dispatch a single message, so we don't really have any way around that sadly.

Your best bet would probably be to maintain a message queue into the `DispatchData`, and accumulate messages from the callbacks during the dispatch (this is for this kind of stuff that SCTK always keeps the `DispatchData` available to all callbacks), and then process the messages from this queue however you want.

Matthias-Fauconneau:
Yeah, I understand that. My argument is that since
prepare_queue/read_events exposes the lock, the same could be done on the
other end.
Yes, RefCell would be a workaround.
Would such an API (disabled by C feature) be in the scope of
wayland-rs/client ?
I need to share mutable state between my dispatch callbacks (e.g add a
keyboard repeat stream).
I feel like adding yet another messaging layer defeat the purpose of using
wayland-rs/client dispatch in the first place.
What would the combined logic actually do ? Sounds like a noop to me to
dispatch to callbacks which immediately reconcentrate to a similar queue
again. Just to get that queue flowing through a stream.

vberger:
> My argument is that since prepare_queue/read_events exposes the lock, the same could be done on the other end.

prepare_read/read_events cannot be called recursively, so there is not problem with them.

> Would such an API (disabled by C feature) be in the scope of wayland-rs/client ?

The problem is, that is not compatible with the way cargo features work: they should only be additive, and never remove functionality. Cargo will automatically enable new features if any crate in the dependency graph needs it, so we can't have a crate relying on a feature being *not* activated.

Matthias-Fauconneau:
The similarity I see is that you can call prepare_read again while already
holding a guard.
Anyway, the solution could be simply to have any reentrant dispatch simply
return immediately since it would have nothing more to dispatch.
Hmm, so it would need to be a forked client crate which doesn't even have a
C feature :/

vberger:
> Or something more like the C API, would be to not have the read_events part borrow the outer event_queue handle, so that dispatch can &mut.

Err, that seems wrong. You're not supposed to call dispatch_pending or dispatch while a `ReadEventsGuard` is in scope. That guard is supposed to be created & consumed on the spot, not kept around. Keeping it around can cause a global deadlock when using the C libs (see #223 ).

Matthias-Fauconneau:
I meant since the prepare_read/read_events combo only needs a connection and dispatch to each objects owner queue, there is no conflict with the specific EventQueue handle used to dispatch.
I can display.create_event_queue a dummy prepare_read/read_events handle which will queue into my main env.queue. 
but maybe I get confused by different semantics of the Rust vs C impl.

vberger:
> I can display.create_event_queue a dummy prepare_read/read_events handle which will queue into my main env.queue.

Yes indeed. The main reason why `prepare_read()` is attached to a queue (rather than to the display for example) is because it checks if the queue has messages that need dispatching and returns `None` if this is the case. This is (yet again) something imposed on us by the C lib API.

I agree this is unfortunate that the lib does not really plays well with async/await, but this is I believe mostly a consequence of the fact that the C libs were absolutely not designed with that in mind. async/await is not exactly a common pattern in GUI programming.

And at the moment, compatibility with the C libs is absolutely required for OpenGL or Vulkan support, this locks us into following their constraints...

PR #318, opened 2020-05-19
Remove dependency on xcur

esposm03:
The `xcur` crate had some bugs, so here is a new implementation which (hopefully) fixes them.
@vberger


vberger:
The PR itself looks good, but I'll need to take some time to test it. :)

repi:
This would be great to get in! 

As then `wayland-cursor` wouldn't depend on a GPL-3.0 crate which would resolve #319

vberger:
Alright, this seems to work well, so lets merge it!

Issue #319, opened 2020-05-21
Replace xcur GPL-3.0 dependency?


repi:
Ran into that the [`xcur`](https://crates.io/crates/xcur) crate that `wayland-cursor` is using is licensed exclusively under GPL-3.0, is that known and intentional that `wayland-cursor` brings in a copyleft license to an otherwise MIT licensed project?

This can probably be a surprise to users, and limit usage in applications that are not licensed under GPL-3.0.

For example in some of our own closed source applications we use (our) [cargo-deny](https://github.com/embarkStudios/cargo-deny) tool to detect incompatible licenses which is how I ran into this:

![image](https://user-images.githubusercontent.com/1262692/82601577-11004080-9bb0-11ea-8f4c-7e22381622d8.png)

Saw that just 2 days ago PR #318 was created specifically removing [`xcur`](https://crates.io/crates/xcur) in favor of [`xcursor`](https://crates.io/crates/xcursor) which is using the permissive MIT license so would resolve this, which would be great!

If you would be open to it, I can PR and a [GitHub Action with cargo-deny](https://github.com/embarkStudios/cargo-deny-action) to verify only licenses you want to support, quite a few repos that use that for this (and more)

vberger:
> If you would be open to it, I can PR and a GitHub Action with cargo-deny to verify only licenses you want to support, quite a few repos that use that for this (and more)

Oh, I did not know about that. That'd be nice indeed!

vberger:
`wayland-cursor` 0.26.5 is now released without `xcur`.

PR #320, opened 2020-05-21
Add cargo-deny to CI to verify licenses and more

repi:
Enabled the current permissive licenses that were used but disallowed copyleft licenses. Issues like #319 should not be able to happen again with this, will be detected in the PR and fail the build.

Also enabled some other standard security settings we usually use such as:
- Verify no dependencies have RUSTSEC advisories for them for vulnerabilities, unmaintained status and general notices
- Disallow bringing in multiple versions of the same crate, to make it leaner and keep compile times good. This can require some tweaking occasionally and in some crate upgrade scenarios temporarily allow ("skip") some crates to have multiple versions of them. 

Feel free to tweak and configure `deny.toml` further, the [cargo-deny book](https://embarkstudios.github.io/cargo-deny/) has quite in-depth descriptions of it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,5 +102,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v1
+      - name: Checkout sources
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
       - uses: EmbarkStudios/cargo-deny-action@v1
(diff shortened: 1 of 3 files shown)


repi:
Hmm there seems to be some bug or mis-configuration here causing cargo-deny to not detect `xcur` and fail here as it should (as this branch is before `xcur` got removed). 

We'll have to debug and investigate further, filed https://github.com/EmbarkStudios/cargo-deny/issues/197

Stupremee:
Don't you fix that by using the `--workspace` argument?

PR #324, opened 2020-05-28
wayland-cursor: don't preallocate user requested size

kchibisov:


wayland-cursor/src/lib.rs
```diff
@@ -95,18 +95,28 @@ impl CursorTheme {
...
+        const INITIAL_POOL_SIZE : i32 = 16 * 16 * 4;
+
+        //  Create shm.
         let mem_fd = create_shm_fd().expect("Shm fd allocation failed");
-        let file = unsafe { File::from_raw_fd(mem_fd) };
-        let pool = shm.create_pool(file.as_raw_fd(), pool_size);
+        let mut file = unsafe { File::from_raw_fd(mem_fd) };
+
+        // Ensure that we have the same we requested.
+        file.write_all(&[0; INITIAL_POOL_SIZE as usize]).expect("Write to shm fd failed");
```
vberger, line 108: You should call `file.flush()` as well, to ensure the compositor has access to the buffer when it tries to map it.

PR #326, opened 2020-05-29
client: server: make WAYLAND_DEBUG more pretty

kchibisov:


wayland-commons/src/wire.rs
```diff
@@ -95,6 +95,21 @@ impl Argument {
...
+impl std::fmt::Display for Argument {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Argument::Int(value) => write!(f, "{}", value),
+            Argument::Uint(value) => write!(f, "{}", value),
+            Argument::Fixed(value) => write!(f, "{}", value),
+            Argument::Str(value) => write!(f, "{:?}", value),
+            Argument::Object(value) => write!(f, "{}", value),
+            Argument::NewId(value) => write!(f, "{}", value),
+            Argument::Array(value) => write!(f, "{:?}", value),
```
kchibisov, line 107: `Str` and `Array` are using `Debug` formatting, since it looks reasonable and doesn't add any complex logic.

vberger:
It looks like a large part of this PR is formatting changes.... weren't they supposedly taken care of by #325 ?

kchibisov:
Well, the split by `\` was done manually, since I've seen the code(done automatically by nightly, but I was on stable), I've also manually formatted some code to make less nested. There're only a few lines which was auto formatted somehow, but CI seems to pass.

wayland-client/src/rust_imp/mod.rs
```diff
@@ -183,3 +196,39 @@ pub(crate) fn default_dispatcher() -> Arc<Mutex<dyn Dispatcher + Send>> {
...
 }
+
+/// Handler for printing events when `WAYLAND_DEBUG` is `true`.
+fn print_event(
+    interface: &str,
+    id: u32,
+    msg_name: &str,
+    request_name: Option<&str>,
+    args: &[Argument],
+) {
```
vberger, line 207: This function can be moved to `wayland-commons`, so that it can be re-used in `wayland-server` as well.

However, I don't really like the way `msg_name` is overloaded like that to mean something different and how the direction of the message (event or request) is inferred from that. I'd rather have an explicit `bool` argument specifying if the message is sent or received, and an other for the `[ZOMBIE]` annotation.

kchibisov:
>This function can be moved to wayland-commons, so that it can be re-used in wayland-server as well.

I've tried to look into that and what wayland-server logs and what wayland-client logs are completely different things I'll investigate how I can move it though.

wayland-commons/src/debug.rs
```diff
@@ -0,0 +1,71 @@
...
+    // Add a new line.
+    eprintln!();
+}
+
+/// Print the send event to stderr in a following format:
+///
+/// [timestamp] -> interface@id.msg_name(args)
+///
+/// If `is_alive` is `false` the `[ZOMBIE]` is added after `id`.
+pub fn print_send_event(
```
kchibisov, line 27: I've have more clear name, please suggest.

wayland-commons/src/debug.rs
```diff
@@ -0,0 +1,71 @@
+//! Debugging helpers to handle `WAYLAND_DEBUG` env variable.
+
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use crate::wire::Argument;
+
+/// Print the dispatched event to stderr in a following format:
+///
+/// [timestamp] <- interface@id.msg_name(args)
+pub fn print_dispatched_event(interface: &str, id: u32, msg_name: &str, args: &[Argument]) {
```
kchibisov, line 10: If you have more clear name, please suggest.

wayland-server/src/rust_imp/resources.rs
```diff
@@ -69,20 +70,25 @@ impl ResourceInner {
...
+            if WAYLAND_DEBUG.load(Ordering::Relaxed) {
+                debug::print_send_event(
                     I::NAME,
                     self.id,
+                    is_alive,
                     self.object.events[msg.opcode as usize].name,
-                    msg.args
+                    &msg.args,
                 );
             }
```
kchibisov, line 90: The client lib has a notion of `zombie` should this one be changed as well? What it's your thoughts on how to handle early return.

kchibisov:
The first commit is the actual change to make WAYLAND_DEBUG nicer,
The second commit format things a bit
The third flattens src/rust_impl/clients.rs in server impl, since it was very nested and hard to read.

kchibisov:
Ok, it seems like one test was failing due to lack of time on CI, when it was started under coverage????

wayland-commons/src/debug.rs
vberger: `print_dispatched_message` would be more appropriate, as the term covers both requests and events

wayland-commons/src/debug.rs
vberger: `print_send_message`

wayland-commons/src/debug.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    eprint!(" -> {}@{}{}.{}", interface, id, if is_alive { "" } else { "[ZOMBIE]" }, msg_name);
+
+    print_args(args);
+
+    // Add a new line.
+    eprintln!();
+}
+
+/// Print arguments without opening/closing bracket.
```
vberger, line 45: Given the code of this function, I think the comment should be "with" and not "without"

wayland-client/src/rust_imp/display.rs
```diff
@@ -21,6 +23,11 @@ pub(crate) struct DisplayInner {
 
 impl DisplayInner {
     pub unsafe fn from_fd(fd: RawFd) -> Result<Arc<DisplayInner>, ConnectError> {
+        if std::env::var_os("WAYLAND_DEBUG").is_some() {
```
vberger, line 26: In order to match the semantics of `libwayland`, we should activate the debugging only if `WAYLAND_DEBUG` is set to either `client` or `1` (an symmetrically, activate it server-side only if it is set to either `server` or `1`).

kchibisov:
Updated first commit on this PR to resolve mentioned issues.

vberger:
Alright, thanks a lot!

PR #328, opened 2020-05-31
sys: use pkg-config for compile time linking, fixes #327

### LINKED ISSUE #327: Use pkg-config to link to libwayland
Currently, non-dlopen native C lib usage fails on FreeBSD because libwayland isn't found (all libs are on a non-default path). I would imagine the issue is even harder to solve on systems that have unusual paths like NixOS :)

You really should use [the pkg-config crate](https://docs.rs/pkg-config) in build.rs to link `libwayland-*`.

`dlib` however currently does `#[link(name = $link)]` which is not great. I guess we're not expecting anyone to *rename* these libraries, so it would be "fine" to keep that and just add pkg-config, but ehhh. `#[link]` is not great.

myfreeweb:


wayland-sys/build.rs
```diff
@@ -0,0 +1,32 @@
...
+    let features = std::env::vars()
+        .filter_map(|(key, _value)| {
+            let feature = key.trim_start_matches("CARGO_FEATURE_");
+            if key.contains("CARGO_FEATURE_") {
+                Some(feature.to_lowercase().replace('_', "-"))
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<_>>();
```
vberger, line 13: Given each feature is checked only once, what's the gain of creating this `Vec` instead of just checking `std::env::var_os("CARGO_FEATURE_FOO").is_some()` at each `if` lower?

myfreeweb: kinda more convenient to check using lowercase names, but sure, let's do the simpler thing

vberger:
Looks good, thanks! :+1: 

Could you just add a changelog entry for this, please?

PR #329, opened 2020-06-04
client: impl clone for QueueToken

kchibisov:
The docs are saying that you can clone the `QueueToken` however it doesn't implement `Clone` so it's either it shouldn't be cloned and docs must be updated or clone should be added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Additions
+
+- [client] Clone implementation for `QueueToken`
+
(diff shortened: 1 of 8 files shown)


kchibisov:
It turned out that rust updated to `1.44.0` and the formatting on stable changed a bit, so the first commit now formats the code to make CI pass.

kchibisov:
And now we have one more commit for clippy, since something became an error right now.

vberger:
Indeed, the missing `Clone` impl was an oversight, thanks!

PR #335, opened 2020-06-11
scanner: parse event type from xml

tadeokondrak:
See https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/82.

Once protocol files are updated, this will remove the need for workarounds to specify these separately.


vberger:
Thanks!

PR #336, opened 2020-06-12
remove byteorder dependency for wayland-client example

JoshMcguigan:
As of Rust 1.32, there is a method in the standard library to convert a u32 to a native endian byte slice. This PR takes advantage of this method to remove the dependency on the `byteorder` crate for the `simple-window` example.


vberger:
Thanks!

PR #337, opened 2020-06-13
modernize simple_window example

JoshMcguigan:
This follows #336 further modernizing the `simple_window` example. 

* remove `extern crate`
* remove `macro_use`
* nested `use` statements
* replace unwraps with `?` by returning `Result` from `main`
* correct or allow all clippy lints

Let me know if there is anything you don't like in here. I also tried to update/extend the comments, but my wayland knowledge is minimal so there wasn't much I could do there. I'd be happy to do this for the remaining examples if there is interest in that.

### DIFF
--- a/wayland-client/examples/simple_window.rs
+++ b/wayland-client/examples/simple_window.rs
@@ -17,8 +17,8 @@ event_enum!(
     Keyboard => wl_keyboard::WlKeyboard
 );
 
-fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let display = Display::connect_to_env()?;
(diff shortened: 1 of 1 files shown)


vberger:
Overall this looks quite good, thanks!

There is just one thing:

- replace unwraps with `?` by returning Result from main

I'd rather keep the unwraps, to keep quite visible in the example the places where something might fail, and may need to be handled.

JoshMcguigan:
Sounds good to me. I've reverted the `?` back to `.unwrap()`.

vberger:
Thanks!

PR #350, opened 2020-09-07
Update core protocol to 1.18

### LINKED ISSUE #348: Update wayland.xml's to 1.18 versions
The current wayland.xml's we're using a targeting 1.16, so those must be updated.

kchibisov:
Fixes #348.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+#### Breaking changes
+
+- [wayland-client] Update core protocol to 1.18.
+- [wayland-server] Update core protocol to 1.18.
+
 #### Additions
 
 - [client] Parsing or protocol errors encountered when reading events are now written to stderr,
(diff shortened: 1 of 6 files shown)


kchibisov:
The xml was taken from https://gitlab.freedesktop.org/wayland/wayland/-/blob/1.18.0/protocol/wayland.xml

kchibisov:
The test change in 0b307ba seems pretty strange, since I don't need it when I run things locally. So maybe something in CI should be changed.

vberger:
> The test change in 0b307ba seems pretty strange, since I don't need it when I run things locally. So maybe something in CI should be changed.

This looks like some rustfmt update again changed the defaults. Maybe we should change the test machinery so that it does not care about that.

Though, given the kind of update this does (it adds new variants to enums), this is a breaking change as far as rust is concerned, the changelog should reflect that.

kchibisov:
I've added changelog entries to the first commit.

I'm still unsure about that test failure wrt rustfmt. Could you repro it locally, since I can't and for me it's the other way around. I've tried both nightly and stable rustfmt.

PR #352, opened 2020-10-10
cursor: fix crash when memfd isn't available

kchibisov:


vberger:
Ah, so plain SHM does not autogrow when writing to it. Thanks!

PR #353, opened 2020-10-10
Use minimal rust version for lints on CI

kchibisov:


vberger:
I prefer to keep the lints on stable channel, so that we can be aware of new lints as they are added, and generally keep an up-to-date code quality.

The cost of one minor commit every 6 weeks to update the few lines that need to be updated seems small enough.

kchibisov:
>The cost of one minor commit every 6 weeks to update the few lines that need to be updated seems small enough.

The point is not to ignore  new clippy updates, but to don't perform silencing of clippy warnings permanently, since the only way to fix is to silence them like you did [here](https://github.com/Smithay/wayland-rs/commit/a5ef9a5c618b673309fece266f66b0d5d0909e45#diff-823cc8d41a370077df8b2fa0d87ca67cb9d64ae48bac41422852881acb38ccc0R87), so you'll never get any new warnings for that check, unless you remember to re enable it when you'll be bumping version, thus you'll be ignoring more than fixing clippy issues.

vberger:
> unless you remember to re enable it when you'll be bumping version

That was my plan, yeah. Though that's a good point, this ignoring would better be done directly in the ci file, rather than in the code like I did. Much less likely to be forgotten.

PR #354, opened 2020-10-17
scanner: automatically use common wayland-rs deps

kchibisov:
wayland-scanner implies the use of wayland-rs, so include
necessary 'use' statements in resulted generated code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Additions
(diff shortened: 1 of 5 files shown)


kchibisov:
it turned out that `wayland-protocol` has a great macro to export all the things, however it's not really easy to find it, unless you explicitly look into `wayland-protocols` docs. And we can't preinclude things, since both server and client share a lot of things.

Issue #349, opened 2020-09-06
Make wayland-rs fully forward compatible


vberger:
The Wayland protocol & wayland-rs are, in general, quite robust to having one side of the socket using a newer version of the protocol than the other.

Wayland-rs has one blind spot on this though: protocol enums. Protocol updates are allowed to add new variants to an enum, with the expectation that clients will generally ignore unknown variants. Wayland-rs currently cannot do this "ignoring", as it translates the values to a Rust enum. Plus, this "ignoring" should probably be handled downstream, not by wayland-rs itself.

PR #355, opened 2020-10-19
Update wlr-protocols

myfreeweb:
foreign-toplevel-management version 3 introduces a new message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
 
 - [cursor] Fix crash on systems without `memfd`.
 
+#### Additions
+
+- [protocols] Update `wlr-protocols` to commit `992841c2aa731ab02774c9bf8c6cda9f24dfbf10`,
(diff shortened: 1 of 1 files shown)


vberger:
Actually, #349 only concerns protocol enums, not the addition of new events or requests. These are already forward-compatible as the associated rust enums are marked `#[non_exhaustive]`. So this alone should not be a breaking change.

Could you add a changelog entry for this update, though?

myfreeweb:
Right, I don't mean breaking in the compile-time sense, I meant that launching the older version on an updated compositor failed with

```
interface 'zwlr_foreign_toplevel_handle_v1' has no event 7
```

…hold on, this means that the server (wlroots) doesn't check the version the client has requested!! not that something is wrong here \*facepalm\* I'll report it there.

> Could you add a changelog entry for this update, though?

Done.

vberger:
Thanks!

PR #356, opened 2020-11-08
cursor: account for both width and height when picking nearest cursor

kchibisov:
Cursors don't have square form in general, so we should match
both width and height when picking nearest images.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Bugfixes
+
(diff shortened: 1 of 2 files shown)


vberger:
Looks good to me, thanks!

Has this been tested on a non-square cursor theme?

kchibisov:
>Has this been tested on a non-square cursor theme?

Yes, on this one https://www.gnome-look.org/p/999983/ . It has pretty weird sizes.

PR #362, opened 2020-12-30
wayland-commons: avoid empty sendmsg on flush

danieldg:
If there is nothing to send, there's no point actually making the system call.
The list of FDs is not checked because sending FDs without also sending data is
not permitted anyway.

### DIFF
--- a/tests/protocol_errors.rs
+++ b/tests/protocol_errors.rs
@@ -38,8 +38,9 @@ fn client_wrong_id() {
 
     server.answer();
 
-    // server should have killed us due to the error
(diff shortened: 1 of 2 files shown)


vberger:
While in principle I agree with this change, it looks like it interferes significantly with the test suite, which uses the return value of flush to detect whether the socket was closed. Introducing this change requires updating the test suite accordingly.

danieldg:
The fact that I had to add handling for the protocol error message in server_native but not without makes me wonder if the rust server is acting correctly on errors.  Regardless, this should fix the tests.

vberger:
Hmm, that's interesting. Maybe the rust server does not properly flush its socket before killing a client?

vberger:
I'll investigate the server sending problem later, thanks!

PR #363, opened 2021-01-15
Unset WAYLAND_SOCKET when we use the socket

danieldg:
Since we are either taking ownership of the socket or closing it at this point,
we should not retain the environment variable as it will only confuse child
processes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


vberger:
Thanks, furthermore it aligns our behavior with libwayland.

LGTM, could you just add a changelog entry for this?

danieldg:
Added

vberger:
Thanks!

PR #364, opened 2021-01-16
Make crates buildable with minimal dependency versions

lo48576:
Some dependencies seems outdated, and does not compile with rustc 1.49.0.
This pull request bumps some dependencies in order to all crates compile with minimal dependency versions.

You can generate `Cargo.lock` with minimal dependency versions by `cargo +nightly update -Z minimal-versions`.
(Once `Cargo.lock` is generated, build and test can be proceeded by stable rustc.)

This PR chose literally minimum compilable deps versions to minimize unexpected side effects, but it results in compiler warnings in some crates (for example, `lazy-static v1.0.2` uses `std::sync::ONCE_INIT` internally, [which is now deprecated](https://doc.rust-lang.org/stable/std/sync/constant.ONCE_INIT.html)).
I think some dependencies should be updated to more recent versions, but it is out of scope of this PR.

### DIFF
--- a/wayland-commons/Cargo.toml
+++ b/wayland-commons/Cargo.toml
@@ -14,5 +14,5 @@ readme = "README.md"
 [dependencies]
 wayland-sys = { version = "0.28.3", path = "../wayland-sys" }
 nix = "0.18"
-once_cell = "1.0"
+once_cell = "1.1"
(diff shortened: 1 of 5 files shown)


vberger:
Thank you!

PR #371, opened 2021-01-30
Update nix dependency to 0.19

chrisduerr:


vberger:
I think the tests are failing because this PR does not update `nix` in the `Cargo.toml` at the workspace root, which is used by the test suite.

vberger:
Thanks!

PR #375, opened 2021-02-14
client: proper cleanup on send on system impl

### LINKED ISSUE #358: Dropping resources kept in filter (quick_assign) move closures
It's very common to keep state in `move` closures `quick_assign`ed to handle events… and to leak it. In fact client-toolkit's `map_keyboard` assigns a big big object that contains `libxkbcommon` state which shows up prominently in memory leak detectors like `heaptrack`.

It doesn't disappear on `Main<T>::release()` or any destructor, it doesn't disappear when the `Main` is dropped (and I haven't created any other handles, so I guess I could say "when all handles are dropped"?), so the only way I found to drop that stuff is to assign a new closure, e.g. `kb.quick_assign(|_,_,_| ())`. Maybe there should at least be an `unassign` method or something.

vberger:
This ensures that the user_data associated with a proxy is
correctly cleaned up when sending a destructor request.

Also, the dispatch logic was updated to correctly handle the case where
an object is destroyed from within its own dispatch closure.

Fixes #358

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,7 @@ jobs:
       - name: Coverage
         uses: actions-rs/tarpaulin@v0.1
         with:
+          version: '0.16.0'
           args: --ignore-tests --features "${{ matrix.features }}" --workspace --exclude wayland-egl --exclude wayland-cursor
       
       - name: Upload to codecov.io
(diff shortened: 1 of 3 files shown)


vberger:
Okay this is disturbing, the coverage part of the CI fails there, while it succeeds when I run it locally.

PR #379, opened 2021-02-26
Upgrade to dlib 0.5

milkey-mouse:
This upgrades wayland-sys' dlib dependency to the new release. Now the dlopen feature won't interfere with other crates using dlib. I have tested this with the `dlopen` feature enabled and disabled on smithay-client-toolkit.

### DIFF
--- a/wayland-sys/src/cursor.rs
+++ b/wayland-sys/src/cursor.rs
@@ -51,9 +51,9 @@ lazy_static::lazy_static!(
                         "libwayland-cursor.so.0"];
 
         for ver in &versions {
-            match WaylandCursor::open(ver) {
+            match unsafe { WaylandCursor::open(ver) } {
                 Ok(h) => return Some(h),
-                Err(::dlib::DlError::NotFound) => continue,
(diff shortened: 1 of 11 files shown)


wayland-sys/src/cursor.rs
```diff
@@ -53,7 +53,7 @@ lazy_static::lazy_static!(
         for ver in &versions {
             match WaylandCursor::open(ver) {
```
vberger, line 54: This call should be unsafe, but the CI didn't catch it because the `cursor` feature of Wayland-sys is no longer used.

wayland-sys/src/egl.rs
```diff
@@ -32,7 +32,7 @@ lazy_static::lazy_static!(
         for ver in &versions {
             match WaylandEgl::open(ver) {
```
vberger, line 33: This call is unsafe as well.

vberger (changes requested):
Thanks, these changes seem good! There are just two `unsafe` block missing.

vberger (approved):

PR #381, opened 2021-03-03
Full rework of the library structure

### LINKED ISSUE #223: Properly implement 'prepare_read' in full-rust wayland-client

### LINKED ISSUE #286: Reduce closures stacking
Following the new `Filter`-based API, several parts of the crate have some cases of closure stacking, where the filter provided by the user is immediately wrapped into a closure before being passed down to the internals of the crate.

In most cases this is unnecessary, and some refactor should allow to get rid of it.

### LINKED ISSUE #288: `offset_of` is indeed UB
There's a `offset_of!` impl here: https://github.com/Smithay/wayland-rs/blob/ff15bdce3410f26d84cdf603b484b3b712bc8620/wayland-sys/src/server.rs#L198

There is no sound way to implement `offset_of!` until [RFC 2582](https://rust-lang.github.io/rfcs/2582-raw-reference-mir-operator.html) lands. The validity rule run afoul of in this case is described in the RFC:

> references must be aligned and dereferencable, even when they are created and never used.

### LINKED ISSUE #317: Crates organization & backend abstraction
I'm re-opening an issue about this to collect thoughts, given a few persons seemed recently interested in improving the API design of wayland-rs (cc @tadeokondrak, @Matthias-Fauconneau).

Currently, wayland-rs is still in tension between two goals:

- designing an API for the Wayland protocol that is as convenient to use as possible from Rust
- keep compatibility with the system C libraries, whose API is not really rust-friendly

The current state is the result of some trade-off between those two, though it's still probably more tailored to the C lib than it should.

So I'd been thinking a few months ago to try an create some kind of "backend abstraction layer", that would be the most minimal & low-level API possible, that would still be implementable both in Rust and on top of the C `libwayland`. From this very low level primitive, it'd then be possible to build an higher-level API that would be more adapted to the Rust language, probably. Though I feel I've kind of exhausted my imagination wrt that for now.

Anyway, this low-level API I imagined looks like that:

```rust
/*
 * Common types
 */

/// Description of the protocol-level information of an object
pub struct ObjectInfo {
    /// The protocol ID
    id: u32,
    /// The interface
    interface: &'static str,
    /// The version
    version: u32,
}

/// A trait representing your data associated to an object
///
/// You will only be given access to it as a `&` reference, so you
/// need to handle interior mutability by yourself.
///
/// The methods of this trait will be invoked internally every time a
/// new object is created to initialize its data.
pub trait ObjectData {
    /// Create a new object data from the parent data
    fn from_parent(my_info: &ObjectInfo, parent_info: &ObjectInfo, parent_data: &Self)
        -> Self;
    /// Create a new object data for an object without parent
    fn from_nothing(my_info: &ObjectInfo) -> Self;
}

// As well as the protocol descriptions for wayland-commons
pub trait Interface { ... }
pub struct Message { ... }
// ....

/// Core trait for representing a Wayland client
///
/// This is the bare minimum API to represent interaction with the
/// Wayland protocol. In particular everything centers around this
/// object representing the connection. Any kind of multi-thread support
/// or dispatching of messages accross several queues must be implemented
/// at a higher level, by sharing this object using `Arc<RwLock<_>>` or
/// something equivalent depending on your goal.
///
/// A well-behaved implementation of this trait should be `Send` as long as
/// its `ObjData` type parameter is.
pub trait WaylandCoreClient<ObjData: ObjectData> {
    type ObjectId: Clone;
  
    /// Initialize the wayland state on a connected unix socket
    fn connect(fd: RawFd) -> Result<Self, ConnectError>;
  
    /// Get the `wl_display` id
    fn display_id(&self) -> ObjectId;
  
    /// Retrieve the last error that occured if the connection
    /// is in an error state
    fn last_error(&self) -> Option<WaylandError>;

    /// Get the connection FD for monitoring using epoll or equivalent
    fn connection_fd(&self) -> RawFd;
    
    /// Create a new client-side object
    ///
    /// If a parent is provided, the `ObjectData` will be inititialized using
    /// `ObjectData::from_parent` with the parent's information. Otherwise, it'll be
    /// initialized using `ObjectData::from_nothing`.
    ///
    /// The version of the created object is chosen as:
    ///
    /// - the provided `version` if it is not `None`
    /// - otherwise the `parent`'s version if it is not `None`
    /// - otherwise `1`
    fn create_object::<I: Interface>(
        &mut self,
        parent: Option<&ObjectId>,
        version: Option<u32>
    ) -> Self::ObjectId;
    
    /// Access the protocol-level information associated with this object
    ///
    /// Returns `None` if the object id is no longer valid.
    fn object_info(&self, id: &Self::ObjectId) -> Option<&ObjectInfo>;
    
    /// Access your `ObjectData` associated with this object
    ///
    /// The object data is initialized by when the
    /// object is created, and you are responsible for any interior mutability.
    ///
    /// Returns `None` if the object is no longer valid.
    fn object_data(&self, id: &Self::ObjectId) -> Option<&ObjData>;
  
    /// Send given message on given object
    ///
    /// If this message should create a new object, you need to first create it
    /// with `create_object()`, and then send it as an argument in the message as
    /// requested by the protocol.
    ///
    /// Fails if the `ObjectId` is no longer valid.
    ///
    /// This method will not validate that the provided message is valid for given
    /// object, you are responsible for checking it.
    fn send_message(&mut self, id: &Self::ObjectId, msg: Message) -> Result<(), ()>;
    
    /// Receive all pending messages from the server
    ///
    /// Your callback will be called iteratively fro each incoming message from
    /// the server.
    /// The function finally returns `Ok(n)` were `n` is the number of messages
    /// dispatched to your callback.
    ///
    /// If a received message should create a new object, this object will be
    /// automatically be created and its `ObjectId` will be given as part of
    /// the `Message`.
    ///
    /// If no message is pending, it'll not block and immediately return `Ok(0)`.
    fn receive_messages<F>(&mut self, callback: F) -> Result<usize, ReceiveError>
    where F: FnMut(ObjectId, Message));
}

pub trait WaylandFFIClient<ObjData: ObjectData>: WaylandCoreClient<ObjData> {
    /// Initialize this connection in guest mode from a shared `wl_display`
    unsafe fn init_from_external_display(*mut wl_display) -> Self;
    
    /// Get the `wl_proxy` associated to a given wayland object
    fn ptr_from_id(&self, id: &ObjectId) -> Option<*mut wl_proxy>;
    
    /// Get an `ObjectId` representing given `wl_proxy`
    ///
    /// Unless this object was previously created by this library, it'll
    /// be considered as "external". As a result it'll not be possible to
    /// access its `ObjData` (`object_data()` will return `None`), and it'll
    /// not be possible to check that it is still valid when sending requests
    /// to it. You are responsible for ensuring this.
    unsafe fn id_from_ptr(&self, ptr: *mut wl_proxy) -> ObjectId;
    
    /// Take control of the given wayland object
    ///
    /// This object will be initialized as if it was created by this library.
    /// As a result its events will arrive in `Self::receive_message` and it'll
    /// be possible to access its `objData`.
    unsafe fn new_id_from_ptr(&mut self, ptr: *mut wl_proxy) -> ObjectId;
}
```

This is the most generic & low-level API I feel confident about implementing. It can probably be tweaked a little depending on the need, but should be a good starting point for thinking about higher-level API design. The higher-level API would internally own an object implementing this API, and delegate to it the low-level interactions with the socket, mostly.

A similar API can certainly be made for server-side, but I've not yet tried to tackle that. I'd rather see if there's something interesting to do client-side first.

### LINKED ISSUE #365: Client::kill with native impl will not supply DispatchData
When I tried to run anvil from smithay in winit mode with Xwayland (with patch in https://github.com/Smithay/smithay/issues/25 applied), I noticed that closing the window will trigger a panic.

Some tracing showed that the issue is caused by https://github.com/Smithay/smithay/blob/4d012e17a0b37cf8fe39530df0948fafa371b026/src/xwayland/xserver.rs#L184. The DispatchData supplied is empty. Further digging shows that in pure Rust impl, the destructors are only called with `Display::flush_clients`, which will fill DispatchData, but for native impl the client destructors are called right away when invoking `Client::kill`, which does not take any DispatchData.

### LINKED ISSUE #376: Short-lived connections are ill-behaved
There has always been general problems around connection cleanup on close, for now I've mostly kept them ignored, as most programs use a single Wayland connection, and the OS cleans up everything on exit.

So, what currently happens is problematic on both the rust impl and the system impl:

On the Rust impl, all proxies keep alive the connection, which contains the object map, whit contains the object dispatchers, which may contain proxies if they where captured in callback closures, causing a reference cycle preventing everything to be cleaned up when the display is dropped, leaking some memory as well as the connection fd

On the system impl, the `wl_display` is destroyed, and libwayland cleans its resources, but doing so does not invoke our rust destructors (which are tasked with free the user_data for example), meaning we also leak some memory, as well as whatever was captured by the callback closures.

I'm not entirely clear on how to fix all that, but this will probably be related to #317

### LINKED ISSUE #383: GlobalManager::list() return tuple is unclear
[`GlobalManager::list()`](https://docs.rs/wayland-client/0.28.5/wayland_client/struct.GlobalManager.html#method.list) return a vector of tuple, while I'm sure I could dig in the code or just guess what is representing this tuple, this could be much clear to have this function return a structure with named field and corresponding documentation.

Could be reuse in [`GlobalEvent::New`](https://docs.rs/wayland-client/0.28.5/wayland_client/enum.GlobalEvent.html)

### LINKED ISSUE #389: wayland-server: Add event loop support for *BSD
`wayland-server/src/rust_imp/event_loop_glue.rs` uses Linux's `epoll` I/O library for event management in the kernel.

`nix::sys::event` can be used to support `kqueue` on *BSD systems, but requires a separate impl. `kqeue` and `epoll` share some common ideas, but are different enough to make the impl of an abstracted common API a little difficult. There are also minor differences even among the *BSDs, but not sure if they will matter for `wayland-server` usage.

What looks like the best way to progress?

Should there be different `event_loop_glue` impls for the supported platforms using different I/O event kernel APIs?
- something like `event_loop_glue/{linux,freebsd,netbsd,openbsd,...}.rs`
- import into `event_loop_glue.rs` based on target OS

What's the best way to abstract the differences to give users of `event_loop_glue` a stable ABI?

### LINKED ISSUE #406: Add `insert` API to `UserDataMap`
Smithay's example compositor Anvil has some code using the `UserDataMap`
struct from wayland-commons, where a value is first inserted and then
immediately read again. Due to the existing API, this is verbose and
requires an `unwrap`:

```rust
state.data_map.insert_if_missing(|| RefCell::new(Data::default()));
let data = state.data_map.get::<RefCell<Data>>().unwrap();
```

Ideally upon insert, a reference to the inserted value would be returned
allowing consumers to directly access the new element without having to
evaluate the safety of unwraps first:

```rust
// Insert if we don't care if it exists already:
let data = state.data_map.insert(RefCell::new(Data::default()));

// Simple get_or_insert:
let data = state.data_map.get::<RefCell<Data>>()
    .unwrap_or_else(|| state.data_map.insert(RefCell::new(Data::default())));
```

While it would be possible to just implement this using a combination of
the existing `insert_if_missing` and `get` method implementations, this
wouldn't be efficient since we have to go through all values just to
retrieve the freshly inserted value again.

A "proper" approach would likely include changing the
`AppendList::append` method to return a reference to the newly inserted
list element, then combine that with a change to the `UserData::set`
method to return a reference to the newly added value. While this would
still require an unwrap in the `UserData::set`, that would be guaranteed
to be safe and hidden from the consumer. It would be a breaking change
to wayland-commons though. Method naming is also a bit awkward since
`insert` usually returns the old value in STD, so maybe that's not the
best name.

I've mentioned this to @vberger on IRC before and it seemed like he was
open to the idea. But both solutions come with some significant
drawbacks, which is why I've decided against just sending my "simple"
solution as a PR.

Do you think this would be worth it? Consumers might appreciate the
convenience but everything considered it's just a single Anvil example
that has spawned that idea.

### LINKED ISSUE #408: Add `Client::post_error` to post errors to the client's `Display`
This was brought up on matrix some time ago by @DemiMarie, so an issue should exist to track that.

vberger:
Addressing #317, and will probably fix a bunch of other issues.

**To do list:**

- [x] commons
  - [x] API definition
  - [x] protocol XML parsing
- [x] rust backend
  - [x] client-side
  - [x] server-side without internal polling
  - [x] server-side with internal polling
- [x] sys backend
  - [x] client-side
  - [x] server-side
  - [x] FFI APIs
- [ ] backend testing
  - [x] basic infrastructure
  - [ ] high test coverage
- [ ] client frontend
  - [x] API definition
  - [x] porting wayland-EGL
  - [x] porting wayland-cursor
- [ ] server frontend
  - [x] API definition
- [ ] scanner
  - [x] Interface gen proc-macro
  - [x] client codegen proc-macro
  - [x] server codegen proc-macro
- [x] wayland-protocols
- [x] frontend testing

Closes #223
Closes #286
Fixes #288
Fixes #317
Fixes #365
Fixes #376
Closes #383
Fixes #389
Closes #406
Fixes #408


wayland-commons/Cargo.toml
```diff
@@ -12,4 +12,7 @@ keywords = ["wayland"]
 readme = "README.md"
 
 [dependencies]
-downcast-rs = "1.1"
\ No newline at end of file
+downcast-rs = "1.1"
+proc-macro2 = "1.0.11"
+quote = "1.0"
+xml-rs = "0.8"
```
jplatte: Your editor seems to not write a final newline consistently.

wayland-scanner/src/interfaces.rs
```diff
@@ -0,0 +1,106 @@
+use proc_macro2::{Ident, Span, TokenStream};
+
+use wayland_commons::scanner::{Interface, Message, Protocol, Type};
+
+use quote::quote;
+
+pub fn generate(protocol: &Protocol) -> TokenStream {
+    let interfaces = protocol.interfaces.iter().map(generate_interface);
+    quote!( #(#interfaces)* )
```
jplatte:
```suggestion
    protocol.interfaces().iter().map(generate_interface).collect()
```

wayland-scanner/src/interfaces.rs
```diff
@@ -0,0 +1,106 @@
...
+    let const_name = Ident::new(
+        &format!("{}_INTERFACE", interface.name.to_ascii_uppercase()),
+        Span::call_site(),
+    );
+    let iface_name = &interface.name;
+    let iface_version = interface.version;
+    let requests = build_messagedesc_list(&interface.requests, &interface.name);
+    let events = build_messagedesc_list(&interface.events, &interface.name);
+
+    quote!(
```
jplatte: `quote!` is usually called with curly braces.

wayland-scanner/src/interfaces.rs
```diff
@@ -0,0 +1,106 @@
...
+pub fn generate(protocol: &Protocol) -> TokenStream {
+    let interfaces = protocol.interfaces.iter().map(generate_interface);
+    quote!( #(#interfaces)* )
+}
+
+fn generate_interface(interface: &Interface) -> TokenStream {
+    let const_name = Ident::new(
+        &format!("{}_INTERFACE", interface.name.to_ascii_uppercase()),
+        Span::call_site(),
+    );
```
jplatte, line 25: Quote contains [`format_ident!`](https://docs.rs/quote/1.0.9/quote/macro.format_ident.html) to simplify the pattern of `Ident::new(&format!("..."), ...)`.

wayland-scanner/src/interfaces.rs
```diff
@@ -0,0 +1,106 @@
...
+    let iface_name = &interface.name;
+    let iface_version = interface.version;
+    let requests = build_messagedesc_list(&interface.requests, &interface.name);
+    let events = build_messagedesc_list(&interface.events, &interface.name);
+
+    quote!(
+        pub static #const_name: wayland_commons::Interface = wayland_commons::Interface {
```
jplatte: Usually in proc-macros you'd use absolute paths (`::wayland_commons::Interface`) but not sure whether that's important here.

wayland-backend/src/rs/server/common_poll.rs
```diff
@@ -0,0 +1,175 @@
...
+            .map_err(Into::into)
+            .map_err(InitError::Io)?;
+
+        #[cfg(any(
+            target_os = "dragonfly",
+            target_os = "freebsd",
+            target_os = "netbsd",
+            target_os = "openbsd"
+        ))]
+        let poll_fd = kqueue().map_err(Into::into).map_err(InitError::Io)?;
```
i509VCB, line 43: macOS implements kqueue: https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/FSEvents_ProgGuide/KernelQueues/KernelQueues.html

Might be worth adding it to the target_os list

wayland-backend/src/rs/server/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+}
+
+downcast_rs::impl_downcast!(sync ClientData<D>);
+
+/// An id of an object on a wayland server.
+#[derive(Clone)]
+pub struct ObjectId {
+    id: u32,
+    serial: u32,
+    client_id: ClientId,
```
PolyMeilex, line 114: Could we add a getter for `client_id` here?
Or directly in `Resource` trait.

I basically need a way to replicate the behavior of `same_client_as`, so I need a way to get `ClientId`.

wayland-server/src/dispatch.rs
```diff
@@ -0,0 +1,121 @@
...
+        new_data
+    }
+
+    fn destroyed(
+        &self,
+        _: wayland_backend::server::ClientId,
+        _: wayland_backend::server::ObjectId,
+    ) {
+        self.udata.object_destroyed()
+    }
```
PolyMeilex, line 120: As discussed on matrix, could we forward `ObjectId` to the `object_destroyed` method

vberger:
My general idea about how to approach modularity with the 0.30 API : https://gist.github.com/vberger/e1b5392419ce4a62cc31c1a0ea00f724

vberger:
Making a note here to not forget about it: I need to adjust the backend API to expose the `prepare_read` API (or something similar), now that I correctly understand its use.

wayland-backend/src/rs/server/mod.rs
vberger: Would a `same_client_as` method on `ObjectId` do the job, instead? Producing a valid `ClientId` without the `Handle` is not going to be possible on the `sys` backend.

PolyMeilex: I haven't needed direct access to client_Id so far, so just a way to replicate `same_client_as` behavior should be enough.

PolyMeilex: Just as a sidenote, while at it, would it be somehow possible to avoid it's current behavior?
```rs
if !(self.is_alive() && other.is_alive()) {
       return false;
}
 ```
 Pattern of cleaning up stuff in Smithay by using `same_client_as` inside destructors is all over the place, and from my testing I'm pretty sure it does not clean up properly because objects have to be alive to compare them.

PR #384, opened 2021-03-28
Use a BufWriter in simple_window example

Stargateur:
This commit reduce by a lot the time to generate the gradient in simple_window example, and show the good practice to use a buffer to write to a file even temporary.

Testing Wayland on a raspberry pi even on release the generation was taking +4s, now it's take 40 ms, I think it's worth it and doesn't add a lot of code complexity as the gradient generation code is not mean to be important in the simple_window example.

I also added some time information allowing the user to see what take time in the example (where a true use case would probably generate the image way faster using GPU for example). Tell me if you think it's not good to add it.

### DIFF
--- a/wayland-client/examples/simple_window.rs
+++ b/wayland-client/examples/simple_window.rs
@@ -2,8 +2,13 @@
 // are reasonable variable names in this domain.
 #![allow(clippy::many_single_char_names)]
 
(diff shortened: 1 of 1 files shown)


vberger:
That seems a nice addition, thanks!

PR #386, opened 2021-06-02
Update wlr-protocols (layer-shell version 4)

unrelentingtech:
layer-shell version 4 introduces a new keyboard interactivity enum.

---

Unfortunately this is truly API breaking…


vberger:
Thanks!

This is indeed an api-breaking change, but these protocols are behind an `unstable_protocols` cargo feature for a reason.

PR #387, opened 2021-06-03
Added to `request_enum` and `event_enum` the possibility to handle attributes

Uniformbuffer3:
Hi,
with some help from the matrix chat, i have added with this tiny change the possibility to use attributes  for `request_enum!` and `event_enum!` macros. The attributes can be added just before the enum name, for example:
```rust
request_enum!(
    #[derive(Debug)]WaylandRequest |
    Compositor => WlCompositor,
    Subcompositor => WlSubcompositor,
    Shm => WlShm,
    Seat => WlSeat
);
```
The macro will copy everything inside `#[...]`, so it will works also for all the the attributes. 
Thanks for the attention and for the effort involved into the project!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+## 0.28.6 -- 2020-06-3
+
+- [client] Added the possibility to handle attributes for `event_enum!` macro.
+- [server] Added the possibility to handle attributes for `request_enum!` macro.
+
 ## 0.28.5 -- 2020-02-26
 
 - [sys] Update `dlib` dependency to v0.5 to match new macro fornat. The `dlopen` feature no longer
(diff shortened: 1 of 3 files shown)


vberger:
Thanks for this PR!

I'd have two comments before merging it:

- could you add a changelog entry for that?
- If it works, I think it'd be better to use the `*` matcher rather than `?`, so that the user could specify multiple attributes.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+## 0.28.6 -- 2020-06-3
```
vberger, line 5: Please keep the changelog entries in the "unreleased section", I'll add the version header when I actually do the release.

Uniformbuffer3: I'm sorry, i didn't understood how Changelog was managed.

PR #388, opened 2021-06-06
client: Fix not found macro

hhirtz:
Using the `crate::` modifier allows users to invoke this macro with no
`use` import, e.g.

    wayland_client::event_enum! { ... }


vberger:
Ah, nice catch, thank. However I believe inside a macro it is `$crate::` that should be used.

Could you also add a changelog entry for this, please?

vberger:
Thanks!

PR #390, opened 2021-06-28
Added `PartialEq` implementation for the `Client` structure.

Uniformbuffer3:
Hi,

this little patch add `PartialEq` implementation for the `Client` structure by just using the already implemented `equals` function. Implementing `PartialEq` allow to integrate `Client` with other standard rust components that rely on that trait.

Thanks for the attention,
Have a good day


vberger:
The PR needs a `cargo fmt`, but other than that this is good, thanks!

Uniformbuffer3:
Yes, i forgot it. Done.

PR #398, opened 2021-08-02
[server] add wl_client_get_credentials

### LINKED ISSUE #397: Add support for `wl_client_get_credentials`

cmeissl:
closes #397

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* [server] Introduce `Client::credentials()` to allow retrieving the credentials of client socket (implements `wl_client_get_credentials`).
+
 ## 0.29.0 -- 2021-07-24
 
 #### Additions
--- a/tests/server_clients.rs
+++ b/tests/server_clients.rs
@@ -98,8 +98,23 @@ fn client_credentials() {
     let server_client = server_client.lock().unwrap().take().unwrap();
     let credentials = server_client.credentials();
     assert!(credentials.is_some());
+    assert_credentials(credentials.unwrap());
+}
 
-    let credentials = credentials.unwrap();
-
+#[cfg(any(not(feature = "server_native"), not(target_os = "freebsd")))]
+fn assert_credentials(credentials: ways::Credentials) {
     assert!(credentials.pid != 0);
+}
+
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* [server] Allow the `Credentials` to be retrieved from a client socket (implements `wl_client_get_credentials`).
```
vberger, line 5:
```suggestion
* [server] Introduce `Client::get_credentials()` to allow retrieving the credentials of client socket (implements `wl_client_get_credentials`).
```

cmeissl: Yeah, that is way better, thanks!

cmeissl:
I have added a test for the `Client::credentials` function, but it fails on older FreeBSD versions.
After some investigation i think we hit this issue here [Bug 246189 - graphics/wayland: fix wl_client_get_credentials](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=246189)
Not sure how to deal with it. 

- Update the wayland version used in the CI, but the version may be not available
- Ignore the test on older versions? But this could lead to bugs in client code i guess

cmeissl:
Hm, disabling the pid assert on older versions should be enough actually.

vberger:
Hmm, so [`std;:os::unix::ucred`](https://doc.rust-lang.org/stable/std/os/unix/ucred/index.html) is unstable and the `nix` implementation only provides `PeerCredentials` for Linux...

I was thinking we could just get the client fd using `wl_client_get_fd` and do the credentials extraction ourselves, but that does not look good actually. So yeah, sounds like we need to disable that test on older FreeBSD I guess..

cmeissl:
I also checked if we could use nix, but as you already wrote getting the PeerCredentials seems to be unsupported on anything than linux. Not sure if it is the best way but I have split the test assert into two versions, in the freebsd version the wayland-server lib version is checked and if it is older than 1.19.0 the assert is ignored.

cmeissl:
Ok, i changed the test to not check the pid on freebsd native lib. I left the check in place for all other cases, even the native check as this should make sure we passed the pointers correctly to the native lib.

PR #400, opened 2021-08-29
Conditionally exclude the socket layer to port to other network stacks

aredridel:
I am working on porting some wayland based code to a new platform (using webassembly and the browser); the network layer is necessarily different, and there's no way CLOEXEC sockets will be supported. I'd like to re-use wayland-commons however, so I'd love to get this conditional compilation included, especially considering wayland-protocols depends on it.

### DIFF
--- a/wayland-commons/Cargo.toml
+++ b/wayland-commons/Cargo.toml
@@ -16,3 +16,7 @@ wayland-sys = { version = "0.29.0", path = "../wayland-sys" }
 nix = "0.22"
 once_cell = "1.1"
 smallvec = "1"
+
+[features]
+default = ["socket"]
+socket = []
--- a/wayland-commons/src/lib.rs
+++ b/wayland-commons/src/lib.rs
@@ -24,6 +24,7 @@ use wayland_sys::common as syscom;
 pub mod debug;
 pub mod filter;
 pub mod map;
+#[cfg(feature = "socket")]
 pub mod socket;
 pub mod user_data;
 pub mod wire;


vberger:
So, there are a few things to keep in mind here:

- First, once #381 is finished (I don't know when), `wayland-commons` will no longer exist has a part of the `wayland-rs` set of crates, meaning the current published 0.29 will effectively go unmaintained.
- Second, when trying to port Wayland-based code to a new platform, before event thinking about cloexec sockets, there is the question of ancillary data. Will your network stack support the passing of file descriptors onto your sockets, like unix sockets do? This mechanism is quite core to the Wayland protocol.

aredridel:
Makes sense! I'm actually mostly after `wayland-protocols` more than `wayland-common`, but there's currently a dependency there.

Yes, the platform will support fd passing.

vberger:
Coming back to this, porting the whole thing to other network stacks would require either making the whole wayland-rs crates generic on the network stack used, or include support for those network stacks directly in wayland-rs. In both cases this is some pretty serious change to the library, not simply adding a cargo feature.

aredridel:
Yeah. I was hoping to build my own network layer at this point, but re-use the protocol definitions.

vberger:
Well, on a positive note, #381 will introduce a `wayland-backend` crate that separates pretty well the protocol definitions from the wire protocol itself, that crate might be a relevant place to merge support for alternative transport layers.

vberger:
Now that the 0.30 rework is merged, I'm going to close this PR. Please see the current master branch as a starting point if you are still interested in porting the wayland stack to other platforms. :slightly_smiling_face:

PR #403, opened 2021-10-01
Remove reference to non-existent function

WhyNotHugo:
The reference function was deleted in b523022401bed89d3d7123d18a17c106e412a6fe

### DIFF
--- a/wayland-client/src/display.rs
+++ b/wayland-client/src/display.rs
@@ -192,8 +192,6 @@ impl Display {
     ///
     /// Will write as many pending requests as possible to the server socket. Never blocks: if not all
     /// requests could be written, will return an io error `WouldBlock`.
-    ///
-    /// This function is identical to `EventQueue::flush`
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
BTW, `EventQueue::flush` seems like a far more logical place for `flush`. I only found `Display::flush` because it was mentioned on IRC.

vberger:
Thanks!

> BTW, EventQueue::flush seems like a far more logical place for flush. I only found Display::flush because it was mentioned on IRC.

Yeah, the current state of `flush()` reflects the internal behavior: it flushes the whole connection, not a single event queue. But I agree it could be relevant to have an `EventQueue::flush()` method that does the same thing I guess.

WhyNotHugo:
> the current state of flush() reflects the internal behavior: it flushes the whole connection, not a single event queue

Ah, that's good to know.

Thanks!

PR #405, opened 2021-10-09
Adding unstable Input method v2 protocol

rano-oss:
The current input method, v1, is very hard to develop for. V2 is much better :smile:


vberger:
Thanks, that looks good!

Could you just add a changelog entry for this addition please?

vberger:
Thanks!

PR #407, opened 2021-10-14
Add KDE server decoration protocol

PolyMeilex:


vberger:
LGTM thanks. Could you just add a changelog entry for that?

PR #410, opened 2021-11-23
backend: ensure client and server types implement same traits

i509VCB:


wayland-backend/tests/rs_sys_impls.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+mod server {
+    #[test]
+    fn test_impls() {
+        // ObjectData
+        assert_impl!(
+            dyn server::ObjectData<()>: downcast_rs::DowncastSync
+        );
+        #[cfg(not(tarpaulin_include))]
```
vberger, line 112: Why are those annotations here?

i509VCB: Does tarpaulin not apply in testing land?

i509VCB: gone

PR #411, opened 2021-11-24
backend, server: document types in rs and sys backends

i509VCB:
Documents a good portion of the rs and sys types in backends. Some parts I don't understand well enough yet to document.

This is a starting point for the wayland-backend docs, I imagine we will introduce some more docs regarding usage in the near future.


wayland-backend/src/protocol.rs
```diff
@@ -2,34 +2,38 @@ use std::{ffi::CString, os::unix::io::RawFd};
...
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub enum ArgumentType {
-    /// i32
+    /// An integer argument. Represented by a [`i32`].
     Int,
-    /// u32
+    /// An unsigned integer argument. Represented by a [`u32`].
     Uint,
-    /// fixed point, 1/256 precision
+    /// A fixed point, 1/256 precision signed floating point number.
```
vberger, line 21:
```suggestion
    /// A signed fixed point number with 1/256 precision
```

If it is a fixed point number, then by definition it is not a floating point number. ^^

wayland-backend/src/protocol.rs
```diff
@@ -39,27 +43,27 @@ impl ArgumentType {
...
 #[allow(clippy::box_vec)]
 pub enum Argument<Id> {
-    /// i32
+    /// An integer argument. Represented by a [`i32`].
     Int(i32),
-    /// u32
+    /// An unsigned integer argument. Represented by a [`u32`].
     Uint(u32),
-    /// fixed point, 1/256 precision
+    /// A fixed point, 1/256 precision signed floating point number.
```
vberger, line 50:
```suggestion
    /// A signed fixed point number with 1/256 precision
```

wayland-backend/src/protocol.rs
```diff
@@ -95,12 +99,21 @@ impl<Id: std::fmt::Display> std::fmt::Display for Argument<Id> {
...
+/// communicate.
 #[derive(Debug)]
 pub struct Interface {
+    /// The name of the interface.
     pub name: &'static str,
+    /// The maximum supported version of the interface.
     pub version: u32,
+    /// A list that describes every request this interface supports.
     pub requests: &'static [MessageDesc],
+    /// A list that describes every event this request supports.
```
vberger, line 114:
```suggestion
    /// A list that describes every event this interface supports.
```

wayland-backend/src/rs/server/common_poll.rs
```diff
@@ -45,6 +49,10 @@ impl<D> Backend<D> {
         Ok(Backend { handle: Handle::new(), poll_fd })
     }
 
+    /// Initializes a connection to a client. The `data` parameter contains data that will be associated with
```
vberger, line 52: Please add an empty line after the first sentence, for better rustdoc rendering

wayland-backend/src/rs/server/common_poll.rs
```diff
@@ -45,6 +49,10 @@ impl<D> Backend<D> {
...
+    /// the client.
+    ///
+    /// The stream must be the receiver end.
```
vberger, line 55: there is no "receiver end" on a unix socket pair, the stream is bidirectional.

wayland-backend/src/rs/server/common_poll.rs
```diff
@@ -88,24 +96,56 @@ impl<D> Backend<D> {
...
     }
 
+    /// Flushes pending events destined for a client.
+    ///
+    /// If no client is specified, all pending events are flushed to all clients.
     pub fn flush(&mut self, client: Option<ClientId>) -> std::io::Result<()> {
         self.handle.flush(client)
     }
 
+    /// Returns a handle which contains the server side state of the backend.
```
vberger, line 106:
```suggestion
    /// Returns a handle which represents the server side state of the backend.
```

wayland-backend/src/rs/server/common_poll.rs
```diff
@@ -88,24 +96,56 @@ impl<D> Backend<D> {
...
 
+    /// Dispatches all pending messages from the specified client.
+    ///
+    /// This method will not block if there are no pending messages.
+    ///
+    /// The provided `data` will be provided to the handler of messages received from the client.
+    ///
+    /// For performance reasons, use of this function should be integrated with an event loop, monitoring the
+    /// file descriptor retrieved by [`Backend::poll_fd`] and only calling this method when messages are
+    /// available.
```
vberger, line 133: For client-specific dispatch, it's the client-specific FD that should be monitored (the one associated to its `UnixStream`).

Also, please add in the documentation that this method may dispatch other clients as well as the one that was requested when using the sys backend.

wayland-backend/src/rs/server/handle.rs
```diff
@@ -177,10 +181,12 @@ impl<D> Handle<D> {
 }
 
 impl<D> Handle<D> {
+    /// Returns information about some object.
     pub fn object_info(&self, id: ObjectId) -> Result<ObjectInfo, InvalidId> {
         self.clients.get_client(id.client_id.clone())?.object_info(id)
     }
 
+    /// Returns the id of the client which created the object.
```
vberger, line 189:
```suggestion
    /// Returns the id of the client which owns the object.
```

wayland-backend/src/rs/server/handle.rs
```diff
@@ -189,15 +195,18 @@ impl<D> Handle<D> {
...
         let client = self.clients.get_client(id)?;
         Ok(client.data.clone())
     }
 
+    /// Returns an iterator over all clients connected to the server.
     pub fn all_clients<'a>(&'a self) -> Box<dyn Iterator<Item = ClientId> + 'a> {
         Box::new(self.clients.all_clients_id())
     }
 
+    /// Returns an iterator over all objects that have been created by a client.
```
vberger, line 209:
```suggestion
    /// Returns an iterator over all objects owned by a client.
```

wayland-backend/src/rs/server/handle.rs
```diff
@@ -242,18 +254,26 @@ impl<D> Handle<D> {
...
     pub fn post_error(&mut self, object_id: ObjectId, error_code: u32, message: CString) {
         if let Ok(client) = self.clients.get_client_mut(object_id.client_id.clone()) {
             client.post_error(object_id, error_code, message)
         }
     }
 
+    /// Kills the connection to a client.
+    ///
+    /// The disconnection reason determines whether the server should simply terminate the connection or post
+    /// an error.
```
vberger, line 267:
```suggestion
    /// The disconnection reason determines the error message that is sent to the client (if any)
```

wayland-backend/src/rs/server/handle.rs
```diff
@@ -263,6 +283,9 @@ impl<D> Handle<D> {
         self.registry.create_global(interface, version, handler, &mut self.clients)
     }
 
+    /// Disables a global object that is currently active.
+    ///
+    /// The global will be removed from clients which have bound the global. New clients will not know of the global.
```
vberger, line 288:
```suggestion
    /// The global removal will be signaled to all currently connected clients. New clients will not know of the global.
```

vberger (changes requested):
Thanks a lot for this first pass on the documentation!

This is mostly pretty good, just had a few remarks. In only made them on the rs-side of the API but they of course apply to the sys side as well.

PR #418, opened 2021-12-12
Switch default socket name to wayland-1

chrisduerr:
It has come to my attention a while ago that it seems to be good pratice
now to use `wayland-1` as the default socket name rather than
`wayland-0`.

This should prevent clients from connecting to `wayland-0` even if
`WAYLAND_DISPLAY` is not set, which is why this change should be
exclusive to the wayland-server part.

### DIFF
--- a/wayland-commons/src/wire.rs
+++ b/wayland-commons/src/wire.rs
@@ -52,7 +52,7 @@ pub enum ArgumentType {
 
 /// Enum of possible argument as recognized by the wire, including values
 #[derive(Clone, PartialEq, Debug)]
-#[allow(clippy::box_vec)]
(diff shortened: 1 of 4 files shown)


vberger:
Thanks!

However, this specific point is already taken care of in #381, and at this point I'd prefer focus on finishing & merging it rather than merging new stuff in master which will induce rebase needs, which is pretty painful on such a massive PR.

chrisduerr:
> However, this specific point is already taken care of in #381, and at this point I'd prefer focus on finishing & merging it rather than merging new stuff in master which will induce rebase needs, which is pretty painful on such a massive PR.

Thanks for letting me know. I've checked the existing PRs but the name didn't tip me off.

PR #426, opened 2021-12-26
client.native: Don't deref potential nullptr

### LINKED ISSUE #425: `SIGSEGV` while formatting protocol error
When running wezterm and during formatting of an error, I'm running into a `SIGSEGV` due to a null pointer being dereferenced in [`native_lib/display.rs:93`](https://github.com/Smithay/wayland-rs/blob/1512878773f1b3fc440dc9f009c0055ca29ffede/wayland-client/src/native_lib/display.rs#L93). It'd probably make sense to somehow catch and gracefully handle this case here.

wayland-client-bug

vberger:
Fixes #425

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Bugfixes
+
+- [client] Fix a case of possible dereferencing of nullptr
+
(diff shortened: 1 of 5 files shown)


athre0z:
Thanks! :)

PR #432, opened 2022-01-03
client/server: explicit DelegateDispatch state

vberger:


i509VCB (approved):
Wonderful changes that make using the new api as a client much nicer

wayland-client/src/event_queue.rs
```diff
@@ -384,7 +392,7 @@ pub trait DelegateDispatch<
     /// The implementation of this function may vary depending on protocol requirements. Typically the client
     /// will respond to the server by sending requests to the proxy.
     fn event(
-        &mut self,
+        state: &mut D,
```
i509VCB, line 395: nit: should this field be called `data` instead as to match the parameters going into the event queue?

PR #439, opened 2022-01-09
backend: Introduce WEnum::into_result

vberger:


wayland-backend/src/protocol.rs
```diff
@@ -242,6 +242,35 @@ pub enum WEnum<T> {
...
+    }
+}
+
+impl<T> WEnum<T> {
+    /// Convert this [`WEnum`] into a result
+    ///
+    /// This can be used to take advantage of the numerous helper methods on [`Result`] if you
+    /// don't plan to handle the unknown case of this enum.
+    #[inline]
+    pub fn into_result(self) -> Result<T, WEnumError> {
```
kchibisov, line 266: Can't it be done via `From` or `Into` traits?

vberger: Well, I can add a `impl<T> From<Wenum<T>> for Result<T, WEnumError>` as well, but I believe that `.into_result()` is much more practical for use in chaining operations than `.into::<Result<T, WEnumError>>()`.

PR #441, opened 2022-01-30
Remove usage eprintln in list_globals example

Eskpil:
Removed usage of eprintln!() in list_globals.rs example in wayland_client. 

Why? I't dosent make sense, it is not a error we are dealing with it is expected output. I changed it to println!() 
because I think that makes way more sense, it is expected output and it is not a error. 

I noticed i changed size of the buffer. I can fix on the next commit.


vberger:
Indeed, the use of `eprintl` here was not appropriate, thanks!

PR #445, opened 2022-02-04
server: fix delegate destructor

vberger:


wayland-server/src/display.rs
```diff
@@ -73,6 +73,10 @@ impl<D: 'static> Display<D> {
     pub fn remove_global(&self, id: GlobalId) {
         self.backend.lock().unwrap().handle().remove_global(id)
     }
+
+    pub fn get_backend(&self) -> &Arc<Mutex<Backend<D>>> {
```
i509VCB, line 77: `backend`

Remember rust generally follows a convention of getters having the same name as their fields.

PR #450, opened 2022-02-25
protocols: introduce misc/virtual-keyboard-unstable-v1

jameschensmith:
[`virtual_keyboard_unstable_v1` protocol][1] provides an application with requests which emulate the behavior of a physical keyboard. I'm not sure how up-to-date this is, but there is a table which lists clients and compositors with the input protocols they support at which version ([view here][2]). There is [an MR to add the protocol to `wayland-protocols`][3], but my only hesitancy is that there is a note that work is ["Stalled because of use case is not clear and technically a virtual seat might be preferred"][4]. In any case, I created this PR because adding this protocol here would replace the [`zwp-virtual-keyboard` crate][5], and any other one-off implementations.

I also cleaned up some formatting 😊 

[1]: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/e279266f714c7122e9ad97d56d047313f78cfdbe/protocol/virtual-keyboard-unstable-v1.xml
[2]: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/39#implementation-matrix
[3]: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/11
[4]: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/39#virtual-keyboard-unstable-v1-1
[5]: https://crates.io/crates/zwp-virtual-keyboard

### DIFF
--- a/wayland-protocols-misc/CHANGELOG.md
+++ b/wayland-protocols-misc/CHANGELOG.md
@@ -1,3 +1,7 @@
 # CHANGELOG: wayland-protocols-misc
 
 ## Unreleased
+
+### Additions
+
+- Introduce protocol `virtual-keyboard-unstable-v1`.
--- a/wayland-protocols-misc/protocols/virtual-keyboard-unstable-v1.xml
+++ b/wayland-protocols-misc/protocols/virtual-keyboard-unstable-v1.xml
@@ -1,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="virtual_keyboard_unstable_v1">
+  <copyright>
+    Copyright © 2008-2011  Kristian Høgsberg
+    Copyright © 2010-2013  Intel Corporation
+    Copyright © 2012-2013  Collabora, Ltd.
+    Copyright © 2018       Purism SPC
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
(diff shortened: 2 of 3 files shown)


wayland-protocols/src/misc.rs
```diff
@@ -84,7 +84,20 @@ pub mod zwp_input_method_v2 {
...
+    //! The virtual keyboard provides an application with requests which emulate
+    //! the behaviour of a physical keyboard.
+    //!
+    //! This interface can be used by clients on its own to provide raw input
+    //! events, or it can accompany the input method protocol.
+
+    wayland_protocol!("./misc/virtual-keyboard-unstable-v1.xml", []);
+}
+
+pub mod server_decoration {
```
jameschensmith, line 100: All I did here was add a space between the module & the brace.

jameschensmith:
@vberger or @i509VCB, would either of you be able to look at this?

vberger:
We've been having some discussion about the `wayland-protocols` crate, and are probably going to split it into multiple crates depending on how "official" the proposed protocols are, allowing to decouple their versioning in consequence.

I don't feel good about having standardized protocol extensions (wayland-protocols) and unstandardized protocols that are possibly temporary and probably never going to be standardized hosted under the same crate with the same versioning guarantees.

jameschensmith:
Fair point & understood. I appreciate your response! Perhaps this PR can be closed and a new one created once all of that is figured out. I'll leave it to you to wrap up this PR. Let me know if there's anything I can do on my end :relaxed: :+1:

jameschensmith:
@vberger, I noticed that the crates had been split. I updated my branch based off the new layout of the codebase. Do you feel better about the protocol being included now?

vberger:
Yes, all is good to me now, thanks!

PR #451, opened 2022-02-27
[feature]Add ext_session_lock to staging protocols

Shinyzenith:


i509VCB:
We should probably stick to release versions of Wayland-protocols, meaning the submodule should be updated to d3249868.

Note that master is 0.30, you'll want to target the 0.29.x branch otherwise.

Shinyzenith:
> We should probably stick to release versions of Wayland-protocols, meaning the submodule should be updated to d3249868.
> 
> Note that master is 0.30, you'll want to target the 0.29.x branch otherwise.

I have updated the submodule.

Shinyzenith:
Should I close this pr and send it to 0.30 branch?

i509VCB:
If you want to use the protocol with the current non alpha crates you'll probably want to make another pull request for those versions.

Shinyzenith:
> If you want to use the protocol with the current non alpha crates you'll probably want to make another pull request for those versions.

Alright, I'll switch to 0.29x as I intend to use this protocol.

PR #454, opened 2022-02-28
protocol: org-kde-kwin-idle-v1 added to misc. (0.30.x)

Shinyzenith:


Drakulix:
Does this belong to misc or should be create a new namespace for kde-protocols?

Shinyzenith:
> Does this belong to misc or should be create a new namespace for kde-protocols?

Probably best to create kde-protocols in my opinion.

Shinyzenith:
Should I go ahead and make it?

Drakulix:
That is something @vberger should likely decide, I just wanted to bring it up, before this gets merged.

kchibisov:
Also, can't we use kde protocols as git submodule, like we do for wlr protocols? And generate all of them?

Shinyzenith:
> Also, can't we use kde protocols as git submodule, like we do for wlr protocols? And generate all of them?

If the idea is allowed by Victor, I'll go ahead and force push that change to this pr.

vberger:
How much of a public api are the KDE protocols considered by KDE?

Shinyzenith:
> How much of a public api are the KDE protocols considered by KDE?

I'm afraid I don't know the answer to that 😔.

vberger:
Basically, my point is that I'm reluctant to integrating them into the `wayland-protocols` crate if there are not at least some stability guarantees on these protocols, which I fear might not be there if they are considered as implementation details by KDE...

vberger:
Closing in favor of #478 which introduces a `wayland-protocols-plasma` crate

PR #458, opened 2022-03-06
backend: allow server to omit ObjectData on error

vberger:
cc @i509VCB


i509VCB (approved):
Testing in smithay confirms this works properly.

PR #468, opened 2022-03-22
Fix segfault during egl surface creation

### LINKED ISSUE #467: Segfault when dropping 0x0 WlEglSurface
When a `WlEglSurface` with the dimensions of 0x0 is dropped, it will cause a segmentation fault. I don't see why this shouldn't be possible, since resizing the surface later is an option, but even if this is explicitly unsupported it should either be marked `unsafe` or cause a panic.

Repro:

```rs
use smithay_client_toolkit as sctk;
use wayland_egl::WlEglSurface;

sctk::default_environment!(Demo);

fn main() {
    let (env, _display, _) = sctk::new_default_environment!(Demo).unwrap();
    let wl_surface = env.create_surface().detach();
    WlEglSurface::new(&wl_surface, 0, 0);
}
```

chrisduerr:
When creating an EGL surface with a width or height of 0, libwayland
will return null as an error. Since `wayland-egl` did not handle this
error, it would lead to a segfault when the surface was dropped.

To resolve this, the surface creation now returns an error, which
indicates that the dimensions are invalid. While using a NonZeroU16
would resolve the need for an error, it comes with a big complexity
overhead for the callee and would restrict some gigantic but possibly
valid surface dimensions.

The only other case where libwayland-egl returns null in the
`wl_egl_window_create` function is when an allocation failure occurs,
which is handled in this patch through a panic. This should also prevent
any other future `null` return values from causing a segfault, though it
might require adjusting the panic message.

Fixes #467.

### DIFF
--- a/wayland-egl/Cargo.toml
+++ b/wayland-egl/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "wayland-egl"
 version = "0.30.0-alpha7"
-documentation = "https://smithay.github.io/wayland-rs/wayland_client/"
+documentation = "https://smithay.github.io/wayland-rs/wayland_egl/"
 repository = "https://github.com/smithay/wayland-rs"
 authors = ["Victor Berger <victor.berger@m4x.org>"]
 license = "MIT"
@@ -14,3 +14,4 @@ readme = "README.md"
 [dependencies]
 wayland-backend = { version = "0.1.0-alpha7", path = "../wayland-backend", features = ["client_system"] }
 wayland-sys = { version = "0.30.0-alpha7", path="../wayland-sys", features = ["egl"] }
+thiserror = "1.0.30"
--- a/wayland-egl/src/lib.rs
+++ b/wayland-egl/src/lib.rs
@@ -12,7 +12,8 @@
 
 use std::os::raw::c_void;
 
-use wayland_backend::{client::InvalidId, sys::client::ObjectId};
+use thiserror::Error as ThisError;
+use wayland_backend::sys::client::ObjectId;
 use wayland_sys::{client::wl_proxy, egl::*, ffi_dispatch};
 
 /// Checks if the wayland-egl lib is available and can be used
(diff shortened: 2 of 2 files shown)


wayland-egl/src/lib.rs
```diff
@@ -40,17 +41,17 @@ impl WlEglSurface {
...
+    pub fn new(surface: ObjectId, width: i32, height: i32) -> Result<WlEglSurface, Error> {
         if surface.interface().name != "wl_surface" {
-            return Err(InvalidId);
+            return Err(Error::InvalidId);
         }
 
         let ptr = surface.as_ptr();
         if ptr.is_null() {
-            Err(InvalidId)
+            Err(Error::NullPointer)
```
vberger, line 51: I think the two checks should be fused into a single if (using `ObjectId::is_null()`), and the `Error::NullPointer` variant can be removed:

```rust
if surface.is_null() || surface.interface().name != "wl_surface" {
    return Err(Error::InvalidId)
}
```

wayland-egl/src/lib.rs
```diff
@@ -112,3 +126,22 @@ impl Drop for WlEglSurface {
...
+    InvalidSize,
+    /// Passed surface object is not a surface.
+    InvalidId,
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        Debug::fmt(self, f)
+    }
+}
```
vberger, line 145: `wayland-client` already depends on `thiserror`, let's use it here as well.

wayland-egl/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "wayland-egl"
 version = "0.30.0-alpha7"
-documentation = "https://smithay.github.io/wayland-rs/wayland_client/"
+documentation = "https://smithay.github.io/wayland-rs/wayland_egl/"
```
chrisduerr, line 4: Unrelated to this PR but really no point opening another PR just to fix this.

PR #478, opened 2022-04-17
Split wayland protocols

vberger:


wayland-protocols/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Breaking Change
+
+- Complete reorganization of the crate around the `wp`/`xdg`/`wp` categories
```
i509VCB, line 7: wp/xdg/ext?

wayland-protocols/src/lib.rs
```diff
@@ -7,27 +7,41 @@
...
-//! The cargo feature `unstable_protocols` adds an `unstable` module, containing bindings
-//! to protocols that are not yet considered stable. As such, no stability guarantee is
-//! given for these protocols.
+//! ## Protocol categories
 //!
-//! Some protocols require unstable rust features, the inclusion of them is controlled
-//! by the cargo feature `nightly`.
+//! The protocols provided in this crate are grouped in 3 main categories:
+//!
+//! - The [`wp`] module contains general purpose wayland protocols
```
i509VCB, line 14: Aren't these plumbing and not general?

wayland-protocols/CHANGELOG.md
HeavyRain266: More like wg/xdg/ext/kde? Or we don't want to bundle kde protocols? or just put them in ext category?

vberger: @i509VCB Indeed, that's a bad copy-paste.

@HeavyRain266 The point of this PR is to split wayland-protocols into multiple crates, so that the `wayland-protocols` crate itself only contains the extensions from the official wayland-protocols upstream repo. Those 3 categories are those from that upstream.

If We are to provide bindings for KDE protocols, they should be maybe in the `wayland-protocols-misc` crate introduced by this PR, or as a `wayland-protocols-kde` crate. However, I'd like to get some clarifications about how much of a public API these protocols are considered by KDE. I'm not willing to take the responsibility of maintaining public bindings to a private API.

wayland-protocols/src/lib.rs
vberger: Well, the wayland-protocols governance document states:

> The "wp" namespace is established for protocols generally useful to Wayland implementations (i.e. "plumbing" protocols).

So, that seems rather correct to me to call them "general purpose", no?

wayland-protocols/CHANGELOG.md
HeavyRain266: I know that from matrix, asked about KDE because there is PR which implements public "idle" protocol used by popular compositors.

vberger: Well, if that protocol is already widely used, it would go in the `wayland-protocols-misc` crate then.

vberger: Ok so, having found the official KDE protocols repo, I'm making a `wayland-protocols-plasma` crate to track it. With a few caveats because some KDE protocols are a little too wonky for `wayland-scanner` to parse.

PR #479, opened 2022-04-24
update to nix 0.24.1

### LINKED ISSUE #457: 0.30 rust backend SIGPIPE signal when client disconnects
At the moment when a client disconnects, the rust backend emits a SIGPIPE signal.

This has no observable effect outside of the debugger since it appears Rust masks sigpipe signals and program execution continues.

There is no way to avoid this at the moment since `ClientData::disconnected` is called after this signal.

i509VCB:
Fixes #457


vberger:
Do you think it could be meaningful to limit nix features, like was done in https://github.com/Smithay/calloop/pull/94 ?

i509VCB:
I'll look into limiting nix features

PR #480, opened 2022-05-07
backend/client: rework API with internal locking in backend

vberger:
This is the result of a discussion on the matrix chatroom about how the previous API was unfriendly to destructors (as the connection handle could not be made available in those contexts).


wayland-backend/src/rs/client_impl/mod.rs
```diff
@@ -772,3 +571,224 @@ impl InnerHandle {
...
+        // If this event is a destructor, destroy the object
+        if message_desc.is_destructor {
+            guard
+                .map
+                .with(message.sender_id, |obj| {
+                    obj.data.server_destroyed = true;
+                    obj.data.client_destroyed = true;
+                })
+                .unwrap();
+            receiver.data.user_data.destroyed(ObjectId {
```
i509VCB, line 769: This may dispatch, causing a deadlock if the user interfaces with the backend again?

wayland-cursor/src/lib.rs
```diff
@@ -160,17 +161,18 @@ impl CursorTheme {
             pool,
             pool_size: INITIAL_POOL_SIZE,
             cursors: Vec::new(),
+            backend: conn.backend().weak(),
```
i509VCB, line 164: To match Rc and Arc, shouldn't that function be called `downgrade` instead?

vberger: I guess :shrug:

wayland-backend/src/rs/client_impl/mod.rs
vberger: If you mean if the user tries to start a dispatch from within a callback, yeah, that would deadlock. You're not meant to do that.

i509VCB: Maybe not dispatch but send a request?

i509VCB: Obviously not to the same object, but a different one

PR #483, opened 2022-05-10
Add 3 new Proxy methods to wayland objects

danieldg:
These are primarily useful for low-level implementations, including those in smithay-client-toolkit, but can also be used by clients if the normal Dispatch mechanism is not a good fit.


wayland-client/src/lib.rs
```diff
@@ -72,6 +82,22 @@ pub trait Proxy: Sized {
...
+    /// Send a request for this object that creates another object.
+    ///
+    /// It is an error to use this function on requests that do not create objects; use
+    /// [Proxy::send_request_nocreate] for such requests.
+    fn send_request_create<I: Proxy>(
+        &self,
+        req: Self::Request,
+        data: Arc<dyn ObjectData>,
+    ) -> Result<I, InvalidId>;
+
```
vberger, line 100: If we want to stick with the conventions from libwayland, we could name them `send_request` and `send_constructor`.

vberger (commented):
In the intent of making it easier to sidestep the `Dispatch` machinery of wayland-rs if needed, I clearly see the interest of those changes. :+1: 

I'm wondering, form a nitpicking point of view, if at this point it would not be relevant to split the `Proxy` trait into two traits, for example `Proxy` and `RawProxy`. The former would include the methods relying on the `Dispatch` machinery, and the later would include the low-level hooks (like the methods added by your PR).

How does that sound?

danieldg:
> I'm wondering, form a nitpicking point of view, if at this point it would not be relevant to split the `Proxy` trait into two traits, for example `Proxy` and `RawProxy`.

Unless we plan to make objects that only implement one of the two traits or there are name collisions, this feels like adding complexity for no real reason.

vberger:
In that case we'll need to make the documentation very clear wrt all those "similar but subtly different methods all at the same place". In my mind, splitting them as two traits would make the API easier to understand for newcomers, for the cost of requiring one more import for people using the low-level API.

danieldg:
I think the only similar-but-different method at the moment is `object_data` vs `data`, and that could be clarified by changing it to `user_data`.  Write vs send is already pretty clear, and `write_request` already has good documentation that it's an internal function.

I'm not sure which methods you were thinking of splitting into `RawProxy` - would that include `parse_event` and `write_request`?

vberger:
> I'm not sure which methods you were thinking of splitting into `RawProxy`

I was thinking about mostly the methods you add in this PR, as they interact directly with the backend-level interfaces rather than being integrated into the `Dispatch<_>` system, but that was not after a lot of thoughts tbh. Let's go with the PR as it is here for now.

wayland-client/src/lib.rs
```diff
@@ -63,6 +64,15 @@ pub trait Proxy: Sized {
     /// Access the user-data associated with this object
     fn data<U: Send + Sync + 'static>(&self) -> Option<&U>;
 
+    /// Access the raw data associated with this object.
+    ///
+    /// For objects created using the scanner-generated interfaces, this will be an instance of the
```
vberger, line 69:
```suggestion
    /// For objects created using the scanner-generated methods, this will be an instance of the
```

Let's write it like this, "interface" is already a pretty overloaded term in wayland context.

PR #484, opened 2022-05-17
server: Allow generics in macros

Drakulix:
I am not sure, if this works exactly in every case, but it appears to be fine? :grimacing:

### DIFF
--- a/wayland-server/src/dispatch.rs
+++ b/wayland-server/src/dispatch.rs
@@ -274,27 +274,30 @@ impl<I: Resource + 'static, U: Send + Sync + 'static, D: Dispatch<I, UserData = 
 /// ```
 #[macro_export]
 macro_rules! delegate_dispatch {
-    ($dispatch_from: ty: [$($interface: ty),*] => $dispatch_to: ty) => {
-        $(
-            impl $crate::Dispatch<$interface> for $dispatch_from {
-                type UserData = <$dispatch_to as $crate::DelegateDispatchBase<$interface>>::UserData;
+    (@impl $dispatch_from:ident $(< $( $lt:tt $( : $clt:tt $(+ $dlt:tt )* )? ),+ >)? : $interface:ty => $dispatch_to: ty) => {
+        impl$(< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $crate::Dispatch<$interface> for $dispatch_from$(< $( $lt ),+ >)? {
+            type UserData = <$dispatch_to as $crate::DelegateDispatchBase<$interface>>::UserData;
 
-                fn request(
-                    &mut self,
-                    client: &$crate::Client,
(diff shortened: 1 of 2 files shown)


wayland-server/src/dispatch.rs
```diff
@@ -274,27 +274,30 @@ impl<I: Resource + 'static, U: Send + Sync + 'static, D: Dispatch<I, UserData =
...
+            fn destroyed(&mut self, client: $crate::backend::ClientId, resource: $crate::backend::ObjectId, data: &Self::UserData) {
+                <$dispatch_to as $crate::DelegateDispatch<$interface, Self>>::destroyed(self, client, resource, data)
             }
+        }
+    };
+    ($impl:tt : [$($interface: ty),*] => $dispatch_to: ty) => {
+        $(
+            $crate::delegate_dispatch!(@impl $impl : $interface => $dispatch_to);
         )*
     };
```
vberger, line 302: Why did you need to do this split here, but not for `delegate_global_dispatch` ?

Drakulix: Oh, that is a good question, maybe we need it in `delegate_global_dispatch` as well and no tests catched it?

For `delegate_dispatch` the compiler was unhappy about unrolling the repetition of `interface` without also unrolling `lt`. I didn't find another way to fix that, than splitting the macro. It was the first doc-test of `delegate_display`, that found this problem (although it appears that it is not even using multiple globals). See here: https://github.com/Smithay/wayland-rs/runs/6475799083?check_suite_focus=true#step:8:289

I have no clue why other tests are working, given that this seems to be quite the fundamental problem, but maybe `delegate_global_dispatch` is lacking a matching test? (I am not that familiar with macros, nor do I have a good understanding of wayland-rs test infrastructure. I just fiddled around with the macro definition until it seemed to work.)

vberger: I think it is that `delegate_global_dispatch` does not have any test yeah, so I believe it'd need the same trick as `delegate_dispatch`.

wayland-server/src/global.rs
```diff
@@ -135,9 +135,9 @@ pub trait DelegateGlobalDispatch<
 
 #[macro_export]
 macro_rules! delegate_global_dispatch {
-    ($dispatch_from: ty: [$($interface: ty),*] => $dispatch_to: ty) => {
+    ($dispatch_from:ident $(< $( $lt:tt $( : $clt:tt $(+ $dlt:tt )* )? ),+ >)? : $interface: ty => $dispatch_to: ty) => {
         $(
```
vberger, line 139: You kept the outer expansion on this macro

Drakulix: :woman_facepalming:

I added a smoke test to actually compile the expanded code from `delegate_global_dispatch` once.

PR #485, opened 2022-05-19
Promote client dispatch data

danieldg:
This promotes the `UserData` associated type to a full generic parameter (on the client only).

### DIFF
--- a/wayland-client/examples/list_globals.rs
+++ b/wayland-client/examples/list_globals.rs
@@ -2,14 +2,12 @@ use wayland_client::{protocol::wl_registry, Connection, Dispatch, QueueHandle};
 
 struct AppData;
 
-impl Dispatch<wl_registry::WlRegistry> for AppData {
-    type UserData = ();
-
+impl Dispatch<wl_registry::WlRegistry, ()> for AppData {
     fn event(
         &mut self,
         _: &wl_registry::WlRegistry,
         event: wl_registry::Event,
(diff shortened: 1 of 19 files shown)


danieldg:
SCTK changes that motivate this: https://github.com/danieldg/smithay-client-toolkit/commits/custom-user-data

vberger:
Ok, with the example it is pretty clear what this aims to achieve. Makes a lot of sense, thanks for designing and proposing this. :+1:

The main design remark I have is: with this change, is the `DelegateDispatchBase` trait still necessary? I had to introduce it to work around complains from rustc wrt generics contraints, maybe with that change we can get rid of it?

danieldg:
> The main design remark I have is: with this change, is the `DelegateDispatchBase` trait still necessary? I had to introduce it to work around complains from rustc wrt generics contraints, maybe with that change we can get rid of it?

It is not, as long as you use the "full" version of `delegate_dispatch` instead of the short one.

vberger:
In that case I'm in favor of removing it.

The pattern we encourage is when providing a `DelegateDispatch` implementation, also provide a macro that expands to the appropriate invocation of `delegate_dispatch!`, so end users will likely never need to invoke `delegate_dispatch!` with user data they didn't create themselves.

vberger:
Also, I'll look to implement something similar server-side as part of #486, thanks again for this design proposal!

PR #486, opened 2022-05-21
backend/server: rework API with internal locking in backend

vberger:


wayland-backend/src/sys/server_impl/mod.rs
```diff
@@ -836,20 +1049,8 @@ impl<D: 'static> InnerHandle<D> {
...
-            return Err(InvalidId);
-        }
-
-        let udata = unsafe {
-            Box::from_raw(ffi_dispatch!(WAYLAND_SERVER_HANDLE, wl_global_get_user_data, id.ptr)
-                as *mut GlobalUserData<D>)
-        };
-        Ok(udata.handler.clone())
+    fn debug(&self, _fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+        todo!()
```
i509VCB, line 1053: Making sure we don't forget.

wayland-server/src/client.rs
```diff
@@ -32,45 +29,49 @@ impl Client {
...
+        let id = handle.handle.create_object::<D>(
+            self.id.clone(),
+            I::interface(),
+            version,
+            Arc::new(ResourceData::<I, _>::new(user_data)),
+        )?;
+        I::from_id(handle, id)
+    }
+
+    pub fn create_resource_objdata<I: Resource + 'static, D: 'static>(
```
i509VCB, line 52: maybe create_with_object_data?

PR #492, opened 2022-05-25
server: expose display ptr on DisplayHandle

i509VCB:


vberger:
I think we should remove the method from `Backend<_>`, to not keep duplicates.

PR #499, opened 2022-06-12
client: add EventQueue::poll_dispatch

danieldg:
This makes it simpler to use multiple event queues with an async runtime for dispatching.


wayland-client/src/event_queue.rs
```diff
@@ -240,6 +244,71 @@ impl<D> EventQueue<D> {
...
+    ///             Either::Right((None, _)) => return Ok(()),
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    pub fn poll_dispatch_pending(
+        &mut self,
+        cx: &mut task::Context,
+        data: &mut D,
+    ) -> task::Poll<Result<Infallible, DispatchError>> {
```
i509VCB, line 297: Is the Ok value supposed to be Infallible or unit?

Unless async runtimes never expect this to return?

danieldg: This particular function will never return an `Ok` variant because you can always keep polling the queue, so `Infallible` is used to denote this.  When using this function, you either need to rely on another future to terminate the task (the end of `app_queue` in the example), or the task will continue running until it gets a `DispatchError` (which might be normal for the "main event queue" of an application)

vberger (changes requested):
LGTM, could you just add an entry in wayland-client's changelog about this?

PR #501, opened 2022-06-17
Link to wlr, plasma and misc protocols in the wayland_protocol docs

hw-lunemann:
Users of releases prior to 0.30.0 won't expect these protocols to be split into their own crates.
This will make it easier for them to find the protocols they are looking for.

### DIFF
--- a/wayland-protocols/src/lib.rs
+++ b/wayland-protocols/src/lib.rs
@@ -34,6 +34,13 @@
 //! This category has now been deprecated and is no longer supposed to be used, however several protocols
 //! are still under that umbrella. We can expect them to be replaced by staging and stable protocols in the
 //! long term, but in the meantime you can enable them with the `unstable` cargo feature.
+//!
+//! ## Other protocols
+//! 
+//! Additionally, more protocol extensions are provided here:
+//! - <https://smithay.github.io/wayland-rs/wayland_protocols_wlr/index.html>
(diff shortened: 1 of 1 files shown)


wayland-protocols/src/lib.rs
```diff
@@ -34,6 +34,13 @@
 //! This category has now been deprecated and is no longer supposed to be used, however several protocols
 //! are still under that umbrella. We can expect them to be replaced by staging and stable protocols in the
 //! long term, but in the meantime you can enable them with the `unstable` cargo feature.
+//!
+//! ## Other protocols
+//!
+//! Additionally, more protocol extensions are provided here:
+//! - <https://smithay.github.io/wayland-rs/wayland_protocols_wlr/index.html>
+//! - <https://smithay.github.io/wayland-rs/wayland_protocols_plasma/index.html>
+//! - <https://smithay.github.io/wayland-rs/wayland_protocols_misc/index.html>
```
vberger, line 43: Rather than raw links to the development docs hosted on smithay.github.io, I think it'd be better to have links with explicit crate names, and pointing to docs.rs, so like:

```
[wayland-protocols-wlr](https://docs.rs/wayland-protocols-wlr)
```

hw-lunemann: I agree, fixed!

vberger (approved):

PR #507, opened 2022-06-24
backend: do not lock state on destruction

cmeissl:
locking the state when invoking the pending
destructors can lead to deadlocks if a
destructor tries to send an event


vberger:
Thanks!

Could you just add an entry for that in `wayland-backend`'s pending changelog?

cmeissl:
> Thanks!
> 
> Could you just add an entry for that in `wayland-backend`'s pending changelog?

Sure, added an entry to the changelog

PR #508, opened 2022-06-25
scanner: derive Ord and Hash for enums

i509VCB:


vberger:
Could you add a changelog entry to wayland-scanner for that?

PR #513, opened 2022-07-01
update wayland.xml

i509VCB:
Wayland 1.21 added some new requests and doc changes


wayland-client/wayland.xml
```diff
@@ -258,6 +258,12 @@
 	for the pool from the file descriptor passed when the pool was
 	created, but using the new size.  This request can only be
 	used to make the pool bigger.
+
+        This request only changes the amount of bytes that are mmapped
```
i509VCB, line 262: Yes upstream had spaces here instead of tabs

PR #514, opened 2022-07-01
backend(client): add Backend::display_ptr

i509VCB:


vberger:
How needed is that given that is the same pointer as the one obtained through `display_id().as_ptr()` ?

i509VCB:
> How needed is that given that is the same pointer as the one obtained through `display_id().as_ptr()` ?

It's less clear imo. It sounds like an implementation detail of libwayland since you'll have to cast the pointer

vberger:
Well, it's part of libwayland's public API that the `wl_display*` is also a `wl_proxy*` (gotta love OOP in C), but yeah, I guess those methods make it clearer.

PR #517, opened 2022-07-03
backend: enable some smallvec features

i509VCB:
Since wayland-rs depends on 1.53, union, const_generics and const_new may be enabled.

### DIFF
--- a/wayland-backend/Cargo.toml
+++ b/wayland-backend/Cargo.toml
@@ -14,10 +14,18 @@ build = "build.rs"
 
 [dependencies]
 wayland-sys = { version = "0.30.0-beta.6", path = "../wayland-sys", features = [] }
-smallvec = "1.4"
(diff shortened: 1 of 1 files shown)


vberger:
What do we gain by enabling those?

i509VCB:
> What do we gain by enabling those?

The union feature decreases the size of a SmallVec by one machine word. So 8 bytes smaller on a 64-bit machine.

vberger:
Ok, but we don't use the `const_generics` nor `const_new`, so why enable them?

i509VCB:
const_generics is for `impl<T, const N: usize> Array for [T; N] { ... }` it replaces the huge macro inside smallvec for the fixed number of array sizes. We don't use an array that large to benefit, but it is more idomatic for the implementation.

const_new probably isnt necessary.

PR #521, opened 2022-07-12
backend/scanner: Fix handling of nullable strings

ids1024:
A `NULL` value in a nullable string is distict from an empty non-`NULL` string. So this changes `Argument::Str` to contain an `Option`, updates `sys` to match how this is represented in libwayland (as a `NULL` pointer), updates `rs` to match the wire protocol for this, and fixes unit tests and `wayland-scanner` accordingly.

Fixes segfault when `NULL` was passed to `wl_data_offer::accept` with the `sys` backend.

I was going to make changes for nullable arrays too, but those appear to be broken in libwayland as well, so that can stay as is for now. (https://gitlab.freedesktop.org/wayland/wayland/-/issues/306)

### DIFF
--- a/wayland-backend/CHANGELOG.md
+++ b/wayland-backend/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - all backends: creating null `ObjectId` is now done through the `ObjectId::null()` method, and the
   `null_id()` methods on the backends are removed.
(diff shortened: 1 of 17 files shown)


vberger:
Oh, thanks! Nice find.

Could you just add an entry to the `wayland-backend` changelog?

ids1024:
Added a mention in the changelog.

vberger:
Thanks a lot!

Issue #421, opened 2021-12-21
Got `NoCompositorListening` from `Display::connect_to_env()`


ArturKovacs:
Using `wayland-client = { version = "0.29", default_features = false,  features = ["use_system_lib"] }`

Various info of my setup:
```
artur@orange:~$ screenfetch 
                          ./+o+-       artur@orange
                  yyyyy- -yyyyyy+      OS: Ubuntu 20.04 focal
               ://+//////-yyyyyyo      Kernel: x86_64 Linux 5.11.0-43-generic
           .++ .:/++++++/-.+sss/`      Uptime: 2h 8m
         .:++o:  /++++++++/:--:/-      Packages: 1875
        o:+o+:++.`..```.-/oo+++++/     Shell: bash 5.0.17
       .:+o:+o/.          `+sssoo+/    Resolution: 1920x1080
  .++/+:+oo+o:`             /sssooo.   DE: GNOME 3.36.5
 /+++//+:`oo+o               /::--:.   WM: Mutter
 \+/+o+++`o++o               ++////.   WM Theme: Adwaita
  .++.o+++oo+:`             /dddhhh.   GTK Theme: Yaru [GTK2/3]
       .+.o+oo:.          `oddhhhh+    Icon Theme: Yaru
        \+.++o+o``-````.:ohdhhhhh+     Font: Ubuntu 11
         `:o+++ `ohhhhhhhhyo++os:      Disk: 52G / 141G (39%)
           .o:`.syhhhhhhh/.oo++o`      CPU: AMD Ryzen 5 4600H with Radeon Graphics @ 12x 3GHz
               /osyyyyyyo++ooo+++/     GPU: AMD RENOIR (DRM 3.40.0, 5.11.0-43-generic, LLVM 12.0.0)
                   ````` +oo+++o\:     RAM: 5642MiB / 7316MiB
                          `oo++.      
artur@orange:~$ mutter --version
mutter 3.36.9
Copyright © 2001-2011 Havoc Pennington, Red Hat, Inc., and others
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
artur@orange:~$ echo $WAYLAND_SOCKET

artur@orange:~$ echo $WAYLAND_DISPLAY
wayland-0

```

vberger:
[...]

PR #524, opened 2022-07-16
Default WAYLAND_DISPLAY to "wayland-0"

Babakinha:
Default WAYLAND_DISPLAY to "wayland-0" instead of giving a NoCompostior error


i509VCB:
I believe this is intentional per this comment: https://github.com/Smithay/wayland-rs/issues/421#issuecomment-1156353764

Babakinha:
> I believe this is intentional per this comment: [#421 (comment)](https://github.com/Smithay/wayland-rs/issues/421#issuecomment-1156353764)

Makes sense

PR #527, opened 2022-07-21
scanner: change OPCODE constants to u16

i509VCB:
This also fixes the event_created_child macro to match a `pat` to allow paths and expressions to work.


vberger (approved):

vberger:
This just need to update the scanner test assets for the CI to pass, otherwise lgtm.

i509VCB:
Have updated the tests

PR #528, opened 2022-07-25
backend/scanner: Do not allow arrays/newids to be nullable

ids1024:
The wire protocol lacked a way to express nullable arrays as distinct from empty non-null arrays, so this was broken in upstream Wayland. Nullable newids were possible in the wire protocol, but not the C API, and dubiously useful.

This matches the change in https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/252.


vberger:
Thanks, this seems pretty good!

Could you just rebase your PR on master? This should take care of CI issues.

vberger:
I see you commited a change to one of the submodules in this PR. I suppose it is a mistake? This is why the CI is failing.

vberger:
Thanks!

PR #530, opened 2022-07-27
client: Add FreezeHandle

danieldg:
This also fixes some potential missed wakeups when using EventQueue methods in a multi-threaded application.

### DIFF
--- a/wayland-client/src/event_queue.rs
+++ b/wayland-client/src/event_queue.rs
@@ -2,10 +2,7 @@ use std::any::Any;
 use std::convert::Infallible;
 use std::marker::PhantomData;
 use std::pin::Pin;
-use std::sync::{
-    atomic::{AtomicBool, Ordering},
-    Arc,
-};
+use std::sync::{atomic::Ordering, Arc};
 use std::task;
 
 use futures_channel::mpsc::{unbounded, UnboundedReceiver, UnboundedSender};
@@ -376,24 +373,20 @@ impl<State> EventQueue<State> {
     /// This function may be useful during initial setup of your app. This function may also be useful
     /// where you need to guarantee all requests prior to calling this function are completed.
     pub fn roundtrip(&mut self, data: &mut State) -> Result<usize, DispatchError> {
-        let done = Arc::new(AtomicBool::new(false));
+        let done = Arc::new(SyncData::default());
 
-        {
-            let display = self.conn.display();
-            let cb_done = done.clone();
-            let sync_data = Arc::new(SyncData { done: cb_done });
-            self.conn
-                .send_request(
-                    &display,
-                    crate::protocol::wl_display::Request::Sync {},
-                    Some(sync_data),
-                )
-                .map_err(|_| WaylandError::Io(Error::EPIPE.into()))?;
-        }
+        let display = self.conn.display();
+        self.conn
+            .send_request(
+                &display,
+                crate::protocol::wl_display::Request::Sync {},
+                Some(done.clone()),
+            )
+            .map_err(|_| WaylandError::Io(Error::EPIPE.into()))?;
 
         let mut dispatched = 0;
 
-        while !done.load(Ordering::Acquire) {
+        while !done.done.load(Ordering::Acquire) {
             dispatched += self.blocking_dispatch(data)?;
         }
 
--- a/wayland-client/src/event_queue.rs
+++ b/wayland-client/src/event_queue.rs
@@ -386,7 +386,7 @@ impl<State> EventQueue<State> {
 
         let mut dispatched = 0;
 
-        while !done.done.load(Ordering::Acquire) {
+        while !done.done.load(Ordering::Relaxed) {
             dispatched += self.blocking_dispatch(data)?;
         }
 
--- a/wayland-client/src/event_queue.rs
+++ b/wayland-client/src/event_queue.rs
@@ -1,12 +1,10 @@
 use std::any::Any;
+use std::collections::VecDeque;
 use std::convert::Infallible;
 use std::marker::PhantomData;
-use std::pin::Pin;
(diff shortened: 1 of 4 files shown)


wayland-client/src/event_queue.rs
```diff
@@ -615,21 +705,35 @@ fn queue_callback<
...
+>(
+    handle: &Connection,
+    msg: Message<ObjectId>,
+    data: &mut State,
+    _: Arc<dyn ObjectData>,
+    qhandle: &QueueHandle<State>,
+) -> Result<(), DispatchError> {
+    let (proxy, event) = I::parse_event(handle, msg)?;
+    if let Some(proxy_data) = proxy.data() {
+        <State as Dispatch<I, U, State>>::event(data, &proxy, event, &proxy_data, handle, qhandle);
```
i509VCB, line 721: Do we need to handle event_created_child here?

wayland-client/src/event_queue.rs
```diff
@@ -496,53 +517,94 @@ impl<State> EventQueue<State> {
...
+
 /// A handle representing an [`EventQueue`], used to assign objects upon creation.
 pub struct QueueHandle<State> {
-    tx: UnboundedSender<QueueEvent<State>>,
+    inner: Arc<Mutex<EventQueueInner<State>>>,
+}
+
+/// A handle that temporarily pauses event processing on an [`EventQueue`].
+#[derive(Debug)]
+pub struct FreezeHandle<'a, State> {
```
i509VCB, line 554: Is there a better name that could be bikeshedded for this type?

danieldg:
See https://github.com/danieldg/smithay-client-toolkit/tree/freeze-popup for an example of use.

wayland-client/src/event_queue.rs
danieldg: No, that's already handled (and is one of the reasons why `make_early_data` needs the type of the userdata even if it doesn't use it directly).

wayland-client/src/event_queue.rs
```diff
@@ -582,6 +622,57 @@ impl<State: 'static> QueueHandle<State> {
...
+        FreezeHandle { qh: self }
+    }
+}
+
+impl<'a, State: 'static> FreezeHandle<'a, State> {
+    /// Create a placeholder object data associated with this event queue.
+    ///
+    /// This is needed when creating composite objects whose user data cannot be produced until all
+    /// components have been created, which would normally produce a cyclic dependency.  If this
+    /// data is not replaced by an object data obtained from [`QueueHandle::make_data`] with
```
vberger, line 650: I'm assuming that "replaced" part is relating to `Backend::set_data`? As discussed on Matrix I'm thinking this method probably should be removed, as it is quite fragile and racy now that the backend has internal locking. So I'm rather uncertain about this.

wayland-client/src/event_queue.rs
```diff
@@ -723,6 +723,8 @@ fn early_callback<
     let (proxy, event) = I::parse_event(handle, msg)?;
     if let Some(proxy_data) = proxy.data() {
         <State as Dispatch<I, U, State>>::event(data, &proxy, event, proxy_data, handle, qhandle);
+    } else {
+        crate::log_debug!("Ignoring event on {}, user_data was never set", I::interface().name);
```
vberger, line 727: This should be *at least* a `log_warning` imo.

vberger (commented):
That's a pretty clean implementation apart from the few nits here and there.

I'm still rather uncertain about the overall proposal, I need to think a bit more about it. Thanks for taking the time to write this as it serves as a solid starting point.

wayland-client/src/event_queue.rs
```diff
@@ -582,6 +622,57 @@ impl<State: 'static> QueueHandle<State> {
 
         Arc::new(QueueProxyData::<I, U> { sender, udata: user_data, _phantom: PhantomData })
     }
+
+    /// Temporarily block processing on this queue.
+    ///
+    /// This will cause the associated queue to appear empty until all [`FreezeHandle`]s associated
+    /// with the queue are dropped.
```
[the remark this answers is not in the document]
danieldg: Yes, this needs to be tweaked; my first pass was as described here, but then I thought about how a dispatch loop with poll would actually work and decided to switch to what is there now.

wayland-client/src/event_queue.rs
danieldg: If `set_data` is removed, then `make_early_data` has no use, but FreezeHandle can still be useful.

wayland-client/src/event_queue.rs
danieldg: It's possible to hit this harmlessly in error conditions (failure to construct a proxy).  With #531 that becomes less of an issue and it may be worth upgrading in severity - or the entire thing becomes moot if dropped.

wayland-client/src/event_queue.rs
vberger: Yeah, I'm thinking we need some kind of mechanism to allow for late-init of the user-data, I'm just wondering if we can manage to do something more robust :thinking:

danieldg:
Marked as draft, should rebase on #535 and possibly remove `make_early_data` depending on the fate of `set_data`.

danieldg:
Removed `make_early_data`, since the primary use for it is no longer necessary due to #531.

wayland-client/src/event_queue.rs
vberger: I'm thinking `QueueFreezeGuard` instead, illustrating that it behaves more like a guard (releasing a resource on drop).

vberger:
Ok so, coming back to this, I get the need for this I can't think of a better way to achieve it, so I plan to merge it. I believe the only remaining nitpick is about the name of the handle/guard?

vberger:
Thanks!

PR #531, opened 2022-07-28
client: Make object creation infallible

danieldg:
Calling methods that do not create objects on a Proxy will ignore `InvalidId` errors.  This extends that functionality to creation methods.

This makes the solution to #530 much simpler (see https://github.com/danieldg/smithay-client-toolkit/commit/inert-proxies) because all failure cases can be handled prior to constructing the proxies.


vberger (commented):
So, this is a big semantic change, but previous iterations of wayland-rs had similar APIs so that's not necessarily wrong.

Just to make sure I understand, is this an alternative to #530 to meet SCTK's needs, or are both PRs needed?

danieldg:
> Just to make sure I understand, is this an alternative to #530 to meet SCTK's needs, or are both PRs needed?

It addresses part of what that PR addresses, and means that the FreezeHandle parts of that PR are not as needed in most situations.  It's still possible to have a fast compositor and a fast other-app-thread conspire to flush, loop, and dispatch an event to a newly created object (XdgSurface, in popup) before `Arc::new_cyclic` increments the strong count, resulting in it looking like an event sent to a dead object - FreezeHandle would prevent this.

vberger (changes requested):
LGTM, thanks.

Could you just add an entry to the wayland-client changelog for that? that's a rather significant breaking change.

PR #532, opened 2022-08-01
feat: generic delegate macros

wash2:
This PR adds support for generics in delegate macros.
Working examples that use the macros are in https://github.com/wash2/client-toolkit


vberger:
Thanks, that looks good! Could you just rebase on master? There is a minor conflict, and that should take care of the CI failure.

vberger:
Thanks!

PR #535, opened 2022-08-07
Improve user data support in ObjectData

danieldg:
This allows using `Proxy::data` on custom `ObjectData` implementations, and simplifies the implementation of QueueProxyData.

### DIFF
--- a/wayland-client/src/event_queue.rs
+++ b/wayland-client/src/event_queue.rs
@@ -1,5 +1,6 @@
 use std::any::Any;
 use std::convert::Infallible;
+use std::marker::PhantomData;
 use std::pin::Pin;
 use std::sync::{
     atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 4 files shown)


wayland-client/src/event_queue.rs
```diff
@@ -602,40 +564,48 @@ fn queue_callback<
...
-        let ret = (self.odata_maker)(&msg);
-        self.sender.send(msg, self.clone());
+        let mut ret = None;
+        for arg in &msg.args {
+            if let Argument::NewId(id) = arg {
+                if !id.is_null() {
+                    ret = Some(State::event_created_child(msg.opcode, &self.handle));
+                }
+            }
+        }
```
i509VCB, line 592: Nitpick here, I imagine this could be rewritten as an iterator where `any` contains a closure which `matches!` for the specific type of argument

i509VCB (approved):

wayland-client/src/event_queue.rs
danieldg: It probably could be turned into a `find_map` call, but I don't think that would be clearer than the loop (and it'd result in the same code in the end, maybe with a `break` after line 590.

i509VCB: Hmm okay, so probably less confusing to leave as is

danieldg: Ah, I think I see what you meant now - `.any(|arg| matches!(arg, Argument::NewId(id) if !id.is_null()))`.  I was trying to assign to `ret` from the iteration.

i509VCB: Yeah I was thinking of something like that

PR #536, opened 2022-08-14
Add feature flag unstable protocols

rano-oss:
I noticed there is no feature flag to enable use of zwp_virtual_keyboard_v1


rano-oss:
I got no clue why this fails, I cannot reproduce it locally

vberger:
I'd say rather the problem here is that that `#![cfg(feature = "unstable_protocols")]` in `wayland-protocols-misc` was a copy-paste error, and should be removed instead.

Regarding the clippy error, it is unrelated to your PR, I just pushed a fix on master.

PR #538, opened 2022-08-18
scanner: migrate to quick-xml

vberger:


jimblandy:
@vberger Thank you for this - it will help `wgpu` clean up our `cargo audit` output once we get a new `wayland-scanner` release, and a `winit` release following that.

When is the next wayland-scanner crate release expected?

vberger:
We are on the final track to release 0.30, you can track what's remaining here: https://github.com/Smithay/wayland-rs/milestone/5

PR #539, opened 2022-08-19
Add a helper to delegate Dispatch<_> to no-op

WhyNotHugo:
I've found this very useful and I'm sure others will too.

It's also useful for interfaces that _don't have_ any events, e.g.:

```rust
// These interfaces have no events, so there's nothing to handle.
dispatch_noop!(State: [wl_compositor::WlCompositor]);
dispatch_noop!(State: [ZwlrScreencopyManagerV1]);
dispatch_noop!(State: [wl_subcompositor::WlSubcompositor]);
dispatch_noop!(State: [wl_subsurface::WlSubsurface]);
dispatch_noop!(State: [wp_viewporter::WpViewporter]);
dispatch_noop!(State: [wp_viewport::WpViewport]);
dispatch_noop!(State: [wl_shm_pool::WlShmPool]);
dispatch_noop!(State: [wl_data_device_manager::WlDataDeviceManager]);
```

### DIFF
--- a/wayland-client/src/event_queue.rs
+++ b/wayland-client/src/event_queue.rs
@@ -791,3 +791,60 @@ macro_rules! delegate_dispatch {
         }
     };
 }
+
+/// A helper macro which delegates a set of [`Dispatch`] implementations for proxies to a static handler.
+///
+/// # Usage
+///
+/// This macro is useful to implement [`Dispatch`] for interfaces where events are unimportant to
+/// the current application and can be ignored.
+///
+/// # Example
+///
+/// ```
+/// use wayland_client::{delegate_noop, protocol::{wl_data_offer, wl_subcompositor}};
+///
+/// /// The application state
+/// struct ExampleApp {
+///     // ...
+/// }
+///
+/// // Ignore all events for this interface:
+/// delegate_noop!(ExampleApp: ignore wl_data_offer::WlDataOffer);
+///
+/// // This interface should not emit events:
+/// delegate_noop!(ExampleApp: wl_subcompositor::WlSubcompositor);
+/// ```
+///
+/// This last example will execute `unreachable!()` if the interface emits any events.
+#[macro_export]
+macro_rules! delegate_noop {
+    ($(@< $( $lt:tt $( : $clt:tt $(+ $dlt:tt )* )? ),+ >)? $dispatch_from: ty : $interface: ty) => {
+        impl$(< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $crate::Dispatch<$interface, ()> for $dispatch_from {
+            fn event(
+                _: &mut Self,
+                _: &$interface,
+                _: <$interface as $crate::Proxy>::Event,
+                _: &(),
+                _: &$crate::Connection,
+                _: &$crate::QueueHandle<Self>,
+            ) {
+                unreachable!();
+            }
+        }
+    };
+
+    ($(@< $( $lt:tt $( : $clt:tt $(+ $dlt:tt )* )? ),+ >)? $dispatch_from: ty : ignore $interface: ty) => {
+        impl$(< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)? $crate::Dispatch<$interface, ()> for $dispatch_from {
+            fn event(
+                _: &mut Self,
+                _: &$interface,
+                _: <$interface as $crate::Proxy>::Event,
+                _: &(),
+                _: &$crate::Connection,
+                _: &$crate::QueueHandle<Self>,
+            ) {
+            }
+        }
+    };
+}


wayland-client/src/event_queue.rs
```diff
@@ -730,3 +730,32 @@ macro_rules! delegate_dispatch {
...
+        $(
+            impl $crate::Dispatch<$interface, ()> for $dispatch_from {
+                fn event(
+                    _: &mut Self,
+                    _: &$interface,
+                    _: <$interface as $crate::Proxy>::Event,
+                    _: &(),
+                    _: &$crate::Connection,
+                    _: &$crate::QueueHandle<Self>,
+                ) {}
```
i509VCB, line 757: The implementation should actually have an `unreachable!()`.

One version of the protocol may have no events but a newer version could.

wayland-client/src/event_queue.rs
```diff
@@ -730,3 +730,32 @@ macro_rules! delegate_dispatch {
...
+/// This macro is useful to implement [`Dispatch`] for interfaces where events are unimportant to
+/// the current application and can be ignored.
+///
+/// ```
+/// use wayland_client::{delegate_noop, protocol::wl_subcompositor};
+/// delegate_noop!(ExampleApp: [wl_subcompositor::WlSubcompositor]);
+/// ```
+#[macro_export]
+macro_rules! delegate_noop {
+    ($dispatch_from: ty: [ $($interface: ty),* $(,)?]) => {
```
i509VCB, line 747: You will need to copy the code from `delegate_dispatch` to also handle generics.

wayland-client/src/event_queue.rs
```diff
@@ -730,3 +730,32 @@ macro_rules! delegate_dispatch {
...
+///
+/// This macro is useful to implement [`Dispatch`] for interfaces where events are unimportant to
+/// the current application and can be ignored.
+///
+/// ```
+/// use wayland_client::{delegate_noop, protocol::wl_subcompositor};
+/// delegate_noop!(ExampleApp: [wl_subcompositor::WlSubcompositor]);
+/// ```
+#[macro_export]
+macro_rules! delegate_noop {
```
i509VCB, line 746: This macro should probably be relocated to `dispatch.rs`

wayland-client/src/event_queue.rs
WhyNotHugo: I also use this macro for interfaces which do have events, but I don't care. For example, an application that can copy, but not paste won't care about data_offer events.

Perhaps two different macros would make sense (one no-op, one unreachable)?

wayland-client/src/event_queue.rs
i509VCB: Actually scratch this, the pull request for that is still pending

wayland-client/src/event_queue.rs
WhyNotHugo: Optionally, the macro could take an `expr` to run, where one could specify `{}` or `unreachable!()` or any other behaviour (e.g.: `println(...`).

wayland-client/src/event_queue.rs
WhyNotHugo: There's no `dispatch.rs` for `wayland_client`, only for `wayland_server`. `delegate_dispatch` is also in this file so this seemed like the best fit.

wayland-client/src/event_queue.rs
```diff
@@ -730,3 +730,47 @@ macro_rules! delegate_dispatch {
...
+            impl $crate::Dispatch<$interface, ()> for $dispatch_from {
+                fn event(
+                    _: &mut Self,
+                    _: &$interface,
+                    _: <$interface as $crate::Proxy>::Event,
+                    _: &(),
+                    _: &$crate::Connection,
+                    _: &$crate::QueueHandle<Self>,
+                ) {
+                    $expr
```
vberger, line 771: I'm not fan of the way this still allows basically any expression to be put here, that's an open road to having unwanted uses of that macro.

Instead and given we'd want to align this on https://github.com/Smithay/wayland-rs/pull/532 , I'd drop the ability to delegate multiple interface in the same call, and change the macro so that its invocation looks something like this:

```rust
// For an interface with no event, examples to `unreachable!()`
delegate_noop!(State: WlCompositor);
// For when you explicitly want to ignore events
delegate_noop!(State: ignore WlCompositor);
```

WhyNotHugo: I like this approach a lot better. Lemme try and figure out how to parse the "ignore" string in a macro rule.

WhyNotHugo: I'm not sure how to pares the `ignore` part of the macro. The Rust docs just say "This section is incomplete" for literals, and I can't find any examples of this being used. Do you know of any? Do you have any pointers for me? I'm kind stuck on that bit.

vberger: AFAIK something like this should work:

```rust
macro_rules! delegate_noop {
    ($dispatch_from: ty : ignore $interface:ty) => {
        /* ... */
    };
    ($dispatch_from: ty : $interface:ty) => {
        /* ... */
    };
}
```

WhyNotHugo: Thanks, that was much simpler than I thought!

I think the latest PR version should match all expectations.

vberger: Not completely, it's still missing the adjustments to allow for generics in the `delegate_to` type, like was done in #532. I'd prefer all such macros follow the same convention. As said earlier, doing this will also mean removing the ability to delegate multiple interfaces in a single macro call.

WhyNotHugo: I don't quite follow, the usage is identical to that of `delegate_dispatch`. Even with #532, you can still pass multiple interfaces to the macro.

WhyNotHugo: Even the example does that: https://github.com/smithay/wayland-rs/blob/8ca0a9b2b390957f15c237bf199b8479148f30ea/wayland-client/src/event_queue.rs#L765

WhyNotHugo: Never mind, the example is broken, test is `ignore`. I'll fix that too.

vberger: Oh yeah, my error, I didn't notice the PR that modified the macro didn't update the example. :(

vberger (approved):

vberger:
@WhyNotHugo the PR as it is looks good to me, are you finished?

wayland-client/src/event_queue.rs
WhyNotHugo: See https://github.com/Smithay/wayland-rs/pull/550

WhyNotHugo:
Yes, I've nothing else to add here.

PR #540, opened 2022-08-20
Simplify ClientData implementation

chrisduerr:
For smithay compositors it is completely optional to implement the
ClientData trait, however due to the required functions on the trait it
is still quite verbose to just implement some dummy struct to use for
it.

Both Smallvill and Anvil have this implemented on a NOOP struct, which
indicates that the majority of compositors will likely not interested in
implementing this themselves.

To make things a little easier for people not interested in implementing
this, a default implementation for `()` is added in this patch. This
allows just passing `Arc::new(())` to `insert_client`.

For people interested in actually using this trait, all functions have
been mode optional. That way it's going to be much cleaner when
compositors just want to provide a partial implementation.


vberger:
LGTM, Thanks!

Could you add an entry to wayland-backend's changelog reflecting this change please?

chrisduerr:
Done!

PR #543, opened 2022-08-21
Android port

lovelyyfiaaa:
This is to supplement https://github.com/Smithay/smithay/pull/711!!

Related PRs that's like this one!! https://github.com/Smithay/calloop/pull/103


wayland-cursor/src/lib.rs
```diff
@@ -375,8 +375,8 @@ pub struct FrameAndDuration {
 
 /// Create a shared file descriptor in memory.
 fn create_shm_fd() -> IoResult<RawFd> {
-    // Only try memfd on linux.
-    #[cfg(target_os = "linux")]
+    // Only try memfd on linux kernel systems.
```
i509VCB, line 378: Nit: linux-like

i509VCB (approved):

wayland-cursor/src/lib.rs
HeavyRain266: FreeBSD also support memfd, though

i509VCB: Hmm didn't know, can you make an issue on sctk about that?

@HeavyRain266 I'd prefer a seperate pull request for that fyi.

lovelyyfiaaa: > Nit: linux-like

Umm... wouldn't Linux-based be better? Since Android's kernel is still based on the kernel.org implementation. 

Though, we could change this to Linux-like when Fuchsia arrives!!

vberger: Well, if it's only about the wording of that comment, I'd suggest "Only try memfd on systems that provide it".

lovelyyfiaaa: that is a good compromise!!

vberger (changes requested):
So, looks like we've settled on how that comment can be updated, and it looks like your patch needs a `cargo fmt` run as the CI is unhappy. Other than that, LGTM, thanks!

lovelyyfiaaa:
> So, looks like we've settled on how that comment can be updated, and it looks like your patch needs a `cargo fmt` run as the CI is unhappy. Other than that, LGTM, thanks!

Okayy!! Done!!

lovelyyfiaaa:
Alsoo, I found some issues with wayland-cursor on the Android target..

```
error[E0432]: unresolved import `nix::sys::memfd`
  --> wayland-cursor/src/lib.rs:61:6
   |
61 | use {nix::sys::memfd, std::ffi::CStr};
   |      ^^^^^^^^^^^^^^^ no `memfd` in `sys`

error[E0425]: cannot find function `shm_open` in module `mman`
   --> wayland-cursor/src/lib.rs:399:21
    |
399 |         match mman::shm_open(
    |                     ^^^^^^^^ not found in `mman`

error[E0425]: cannot find function `shm_unlink` in module `mman`
   --> wayland-cursor/src/lib.rs:407:35
    |
407 |             Ok(fd) => match mman::shm_unlink(mem_file_handle.as_str()) {
    |                                   ^^^^^^^^^^ not found in `mman`

Some errors have detailed explanations: E0425, E0432.
For more information about an error, try `rustc --explain E0425`.
error: could not compile `wayland-cursor` due to 3 previous errors
```

I intended this PR for the Android backend in Smithay, which doesn't depend on wayland-scanner, but what do you think?

vberger:
Well, I suppose wayland-cursor cannot work on android as is. Though I'm not sure the system hierarchy it depends on (the xcursor files and such) would exist on android anyway, so... probably not a big deal. At least I don't think this should block this PR.

lovelyyfiaaa:
Okayy!! Let's get this merged!!

PR #546, opened 2022-08-25
Update `nix` to 0.24.1

### LINKED ISSUE #457: 0.30 rust backend SIGPIPE signal when client disconnects

### LINKED ISSUE #544: Backport `nix` update to 0.24 for `wayland-*` 0.29
Would a PR to update `nix` to 0.24 for `wayland-*` 0.29 be acceptable?

daxpedda:
I fixed #457 for `wayland-*` 0.29 along the way.

Fixes #544.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: ['1.41.0', 'stable', 'beta']
+        rust: ['1.46.0', 'stable', 'beta']
         features: ['', 'client_native', 'server_native', 'both_native']
       
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


daxpedda:
I don't know exactly what the MSRV policy is, can we bump this in a minor version? If that's not possible, then we can close the PR I guess.

vberger:
Generally our policy has been "bumping MSRV is a breaking change".

But considering that it would bump it to 1.43.0 (I think), which is still more than 2 years old, and it's likely going to be the last release on the 0.29.x group (no new features are coming to 0.29), I think we're mostly fine.

Could you just update the CI script to ensure 1.43.0 does work as expected?

daxpedda:
It's 1.46 I believe: https://github.com/nix-rust/nix/tree/r0.24#minimum-supported-rust-version-msrv.

vberger:
Ah, so there are other features than "attributes on if expressions" then. I'd say 1.46.0 is still old enough though.

daxpedda:
Done.

I found:
```rust
// disable clippy lints that are not compatible with rust 1.41
#![allow(clippy::match_like_matches_macro)]
```
Should I fix this while I'm at it?

vberger:
No, I don't think that's necessary. As said the 0.29.x branch is just in maintenance mode while 0.30.0 is being polished.

I'll merge this and make a 0.29.5 release.

PR #551, opened 2022-08-26
client: helper for global initialization

i509VCB:


wayland-client/src/global.rs
```diff
@@ -0,0 +1,226 @@
...
+        U: Send + Sync + 'static,
+    {
+        let version_start = *version.start();
+        let version_end = *version.end();
+        let interface = I::interface();
+
+        if *version.end() > interface.version {
+            // This is a panic because it's a compile-time programmer error, not a runtime error.
+            panic!("Maximum version ({}) of {} was higher than the proxy's maximum version ({}); outdated wayland XML files?",
+                version.end(), interface.name, interface.version);
```
kchibisov, line 116: Isn't it breaking forward compat? You can generate code for older version and then new gets released?

i509VCB: This is from some sctk code. The intention is to prevent programmer error by requesting for example version 5 of zwp_linux_dmabuf_v1 when there is only version 4 right now and then being shot in the foot when version 5 is actually released.

If this is regarding testing new protocol versions, you'll likely need to fork wayland-client/wayland-protocols anyways.

vberger: I don't think this can break forwards compat, given this only panics if you're requesting a higher version than what wayland-rs supports. Updating wayland-rs to a higher supported version can thus never break code.

vberger:
So, this is a starting point that is in the direction of what I had in mind during the discussion about this the other day, but IMO that's missing some important piece: how does that handle globals that are created or removed dynamically during the run of the program?

I had in mind (and planned to work on it) some system where the registry starts out like you wrote here, and then once the initial setup is done switches back to regular `Dispatch<_>` handling of events. But actually I think I can build that on top of your PR here.

vberger:
superseded by #554

PR #554, opened 2022-09-02
client: introduce globals init helpers

vberger:
Based on #551 but allows for both initial global list and handling of dynamic globals

cc @i509VCB

### DIFF
--- a/wayland-client/src/globals.rs
+++ b/wayland-client/src/globals.rs
@@ -1,0 +1,317 @@
+//! Helpers for handling the initialization of an app
+//!
+//! At the startup of your Wayland app, the initial step is generally to retrieve the list of globals
+//! advertized by the compositor from the registry. Using the [`Dispatch`] mechanism for this task can be
+//! very unpractical, this is why this module provides a special helper for handling the registry.
+//!
+//! The entry point of this helper is the [`registry_queue_init`] function. Given a reference to your
+//! [`Connection`] it will create an [`EventQueue`], retrieve the initial list of globals, and register a
+//! handler using your provided `Dispatch<WlRegistry,_>` implementation for handling dynamic registry events.
+//!
+//! ## Example
+//!
+//! ```no_run
+//! use wayland_client::{
+//!     Connection, Dispatch, QueueHandle,
+//!     globals::{registry_queue_init, Global, GlobalListContents},
+//!     protocol::{wl_registry, wl_compositor},
+//! };
+//! # use std::sync::Mutex;
+//! # struct State;
+//!
+//! // You need to provide a Dispatch<WlRegistry, GlobalListContents> impl for your app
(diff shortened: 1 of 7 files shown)


wayland-client/src/globals.rs
```diff
@@ -0,0 +1,291 @@
...
+};
+
+/// Initialize a new event queue with its associated registry and retrieve the initial list of globals
+///
+/// See [the module level documentation](self) for more.
+pub fn registry_queue_init<State>(
+    conn: &Connection,
+) -> Result<(GlobalList, EventQueue<State>), GlobalError>
+where
+    State: Dispatch<wl_registry::WlRegistry, Mutex<Vec<Global>>> + 'static,
```
i509VCB, line 80: Feels a bit raw to expose `Mutex<Vec<Global>>` as the user data type. Do we want to provide a wrapper type that is a little nicer to work with?

wayland-client/src/globals.rs
```diff
@@ -0,0 +1,291 @@
...
+    where
+        I: Proxy + 'static,
+        State: Dispatch<I, U> + 'static,
+        U: Send + Sync + 'static,
+    {
+        let version_start = *version.start();
+        let version_end = *version.end();
+        let interface = I::interface();
+
+        if *version.end() > interface.version {
```
i509VCB, line 149: I'd agree with the coverage warning and I do think we should test this does indeed panic given an invalid version. I guess bind wl_compositor with a version of `u32::MAX`

wayland-client/src/globals.rs
vberger: I did wonder about that, but didn't have the motivation to do it before having your first opinion over the overall API. Will do. :+1:

danieldg:
It would be useful to be able to do the initial round-trip in a non-blocking manner - the existing `registry_queue_init` should remain for the majority of users, but an alternate version that acts like the (current) SCTK RegistryState would be good for applications that can get started on more initial loading while they wait for a round-trip to the compositor.

Also, nitpicking: there are still a few references (docs) to `Mutex<Vec<Global>>` instead of `GlobalListContents`.

vberger:
> but an alternate version that acts like the (current) SCTK RegistryState would be good for applications that can get started on more initial loading while they wait for a round-trip to the compositor.

I'm a little skeptical on what an app would gain from that tbqh. We are at the beginning of the initialization, nothing has been sent on the socket yet, that blocking roundtrip will most likely last for maybe a few milliseconds. Is there really something significant to gain by doing stuff in parallel with that?

I mean yes, we can do an API that'd allow it. But I'm not sure at what would actually be gained from that, while introducing a much less intuitive API in the process.

vberger:
I'm going to merge this as it is now to release the `beta.10`, if we add a non-blocking version of this API this can be done as a separate PR.

PR #558, opened 2022-09-04
Update wlr-protocols submodule

hw0lff:
In the output-management protocol were release requests for `head` and `mode` added.


vberger:
Thanks!

PR #559, opened 2022-09-05
Update plasma-wayland-protocols submodule to v1.7.0

hw0lff:


vberger:
Ok, it looks like some of the XML files were renamed in the `plasma-wayland-protocols` repo, meaning the generation by wayland-scanner now fails because the filenames are not correct any more.

vberger:
Thanks!

PR #560, opened 2022-09-06
client/server: Add Weak helpers

vberger:


wayland-scanner/src/client_gen.rs
```diff
@@ -81,6 +81,12 @@ fn generate_objects_for(interface: &Interface) -> TokenStream {
 
             impl std::cmp::Eq for #iface_name {}
 
+            impl PartialEq<Weak<#iface_name>> for #iface_name {
```
i509VCB, line 84: Is there a reason there can't be a blanket implementation for this?

wayland-client/src/lib.rs
```diff
@@ -234,6 +234,15 @@ pub trait Proxy: Clone + std::fmt::Debug + Sized {
     /// The version of this object
     fn version(&self) -> u32;
 
+    /// Checks if the Wayland object associated with this proxy is still alive
+    fn is_alive(&self) -> bool {
```
i509VCB, line 238: By virtue of having weak proxy/resource wrappers, I'm assuming we needed to readd this.

wayland-scanner/src/client_gen.rs
vberger: Yeah, that conflicts with orphan rules.

wayland-client/src/lib.rs
vberger: Well, that was not strictly needed, but I realized all pieces were already in place to implement this easily so, might as well do it.

PR #572, opened 2022-10-05
client: Add a GTK example

ids1024:
This provides an example of how to use `wayland_client` with a foreign display, in particular GTK. If there is a better way to use `wayland_client` with GTK, the example can be improved.

I'm not sure if this use of `prepare_read` is correct. I guess it can't really run into issues given no other thread is actually reading from the connection? Should it instead create a guard before calling `unix_fd_add_local`, and replace it with a new guard every time the function is called?

### DIFF
--- a/wayland-client/Cargo.toml
+++ b/wayland-client/Cargo.toml
@@ -25,3 +25,6 @@ log = { version = "0.4", optional = true }
 wayland-protocols = { path = "../wayland-protocols", features = ["client"] }
 futures-util = "0.3"
 tempfile = "3.2"
+gtk4 = { version = "0.4.8" }
+gdk4-wayland = { version = "0.4.8", features = ["wayland_crate", "v4_4", "egl"] }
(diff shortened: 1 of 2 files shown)


secext2022:
This example no longer work.

After modify a little to compile, it run with crash:

```sh
thread 'main' panicked at tmp/src/main.rs:81:35:
called `Option::unwrap()` on a `None` value
```

panic at this line:

```rust
    glib::source::unix_fd_add_local(fd, glib::IOCondition::IN, move |_, _| {
        connection.prepare_read().unwrap().read().unwrap();  // !! PANIC !!
        ControlFlow::Continue
    });
```

secext2022:
This is the new working version:

```rust
    glib::source::unix_fd_add_local(fd, glib::IOCondition::IN, move |_, _| {
        if let Some(g) = connection.prepare_read() {
            g.read().unwrap();
        } else {
            connection.backend().dispatch_inner_queue().unwrap();
        }
        ControlFlow::Continue
    });
```

PR #576, opened 2022-11-07
backend: implement HasRawDisplayHandle for Backend in sys

i509VCB:
This makes using wayland-backend within the raw-window-handle ecosystem easier. A new `raw-window-handle` feature has been added to enable this trait implementation


elinorbgr:
LGTM, think you can rebase that PR?

PR #577, opened 2022-11-07
Update wayland-protocols to 1.29

i509VCB:
Since there are now two xwayland protocols, there is now an `xwayland` module for this protocols.

~~Pending on a new wayland-protocols release with https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/174~~


elinorbgr:
Welp, clippy introduced new lints again? 😩

PR #579, opened 2022-11-17
backend: Return `MissingData` instead of `Malformed` if `len > raw.len()`

ids1024:
This allows the higher level code to try reading again to get any remaining bytes of the message, instead of treating the message as malformed.

This matches the implementation of `queue_event()` in libwayland which returns `0` in this case (rather than `-1` for an error).

### DIFF
--- a/wayland-backend/CHANGELOG.md
+++ b/wayland-backend/CHANGELOG.md
@@ -1,8 +1,11 @@
 # CHANGELOG: wayland-backend
 
-## 0.1.0-beta.10
-
 ## Unreleased
+
+#### Bugfixes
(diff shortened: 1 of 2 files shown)


elinorbgr:
LGTM.

elinorbgr:
Could you add a changelog entry for that change please?

elinorbgr:
Uh, looks like the changelog was messed up... The `Unreleased` header is not at its proper place. The entry you added is mixed with the beta.10 changes. Can you fix that maybe? :sweat_smile:

elinorbgr (approved):

PR #583, opened 2022-12-03
Add WaylandSource behind the `calloop` cargo feature

hw0lff:
I have added the `WaylandSource` struct from sctk to the wayland-client crate.
It can be enabled with the `calloop` feature flag.
I also wrote a little description for the feature in `src/lib.rs`.

### DIFF
--- a/wayland-client/src/calloop.rs
+++ b/wayland-client/src/calloop.rs
@@ -1,0 +1,223 @@
+//! Utilities for using an [`EventQueue`] from wayland-client with an event loop that performs polling with
+//! [`calloop`](https://crates.io/crates/calloop).
+
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![warn(missing_docs, missing_debug_implementations)]
+#![forbid(unsafe_op_in_unsafe_fn)]
+#![allow(clippy::new_without_default)]
+
+use std::{
+    io,
+    os::unix::io::{AsRawFd, RawFd},
+};
+
+use crate::{DispatchError, EventQueue};
+use calloop::{
+    generic::Generic, EventSource, InsertError, Interest, LoopHandle, Mode, Poll, PostAction,
+    Readiness, RegistrationToken, Token, TokenFactory,
+};
+use nix::errno::Errno;
+use wayland_backend::client::{ReadEventsGuard, WaylandError};
+
+/// An adapter to insert an [`EventQueue`] into a calloop [`EventLoop`](calloop::EventLoop).
+///
+/// This type implements [`EventSource`] which generates an event whenever events on the display need to be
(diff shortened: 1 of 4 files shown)


wayland-client/src/calloop.rs
```diff
@@ -0,0 +1,223 @@
+//! Utilities for using an [`EventQueue`] from wayland-client with an event loop that performs polling with
+//! [`calloop`](https://crates.io/crates/calloop).
+
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![warn(missing_docs, missing_debug_implementations)]
+#![forbid(unsafe_op_in_unsafe_fn)]
+#![allow(clippy::new_without_default)]
```
elinorbgr, line 7: Please don't add such attributes in the module like this, they are either redundant with the crate-level ones or have no reason to be here.

wayland-client/src/calloop.rs
```diff
@@ -0,0 +1,223 @@
...
+use calloop::{
+    generic::Generic, EventSource, InsertError, Interest, LoopHandle, Mode, Poll, PostAction,
+    Readiness, RegistrationToken, Token, TokenFactory,
+};
+use nix::errno::Errno;
+use wayland_backend::client::{ReadEventsGuard, WaylandError};
+
+/// An adapter to insert an [`EventQueue`] into a calloop [`EventLoop`](calloop::EventLoop).
+///
+/// This type implements [`EventSource`] which generates an event whenever events on the display need to be
```
elinorbgr, line 24:
```suggestion
/// This type implements [`EventSource`] which generates an event whenever events on the event queue need to be
```

elinorbgr (changes requested):
LGTM overall, just a few points to fix.

Also, could you add an entry in the `Unreleased` section of the changelog for wayland-client introducing this change?

wayland-client/src/calloop.rs
hw0lff: Ah, sorry I must have copied the file from the wrong subdirectory. I think this is a leftover from the attempted separate `WaylandSource` crate. I'll remove it!

elinorbgr:
Ok, LGTM, however looks like we have a few CI issues. These are not directly related with your PR, but I plan to fix them before merging.

hw0lff:
What about the renaming of `display` to `event queue`?
Is it only correct in the first instance or all of them?
The topics around it seem to be the same to me.

elinorbgr:
Oh right. Only the changes in the doc comment should be kept.

hw0lff:
Like this [90cbe6f](https://github.com/Smithay/wayland-rs/pull/583/commits/90cbe6ff55cd546abe2608ecbd9f581d537a6475)?

elinorbgr:
Yes 👍

elinorbgr (approved):

PR #586, opened 2022-12-06
Reduce compile time

bjorn3:
The commits are mostly independent, so if you think any of the commits is bad for maintainability, I can easily drop it. The commits I'm most confident about are:

* https://github.com/Smithay/wayland-rs/commit/49779784766a15d937b098289124d4aaffb754b5 "Remove futures dependencies from wayland-client"
* https://github.com/Smithay/wayland-rs/commit/84aad31150043622a5ffd42c9c9a71ec96e2d98f "Reduce code bloat of smallvec![] macro"

### DIFF
--- a/wayland-scanner/src/lib.rs
+++ b/wayland-scanner/src/lib.rs
@@ -38,8 +38,6 @@
 
 use std::{ffi::OsString, path::PathBuf};
 
-use syn::{parse_macro_input, LitStr};
-
 mod c_interfaces;
 mod client_gen;
 mod common;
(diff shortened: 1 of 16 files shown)


wayland-scanner/src/lib.rs
```diff
@@ -49,10 +47,197 @@ mod protocol;
 mod server_gen;
 mod util;
 
+// `bytes`, `next_chr`, `parse_lit_str`, `parse_lit_str_cooked` and `parse_lit_str_raw` are adapted
+// from syn:
+// https://github.com/dtolnay/syn/blob/362ee2d02df3f1b2e74c7b7a4cf2ed3c106404c9/src/lit.rs#L1062-L1167
+// and
+// https://github.com/dtolnay/syn/blob/362ee2d02df3f1b2e74c7b7a4cf2ed3c106404c9/src/lit.rs#L1327-L1388
```
vberger, line 54: Could you instead put all this parsing code into its own file, to avoid overloading hte `lib.rs` file?

vberger (changes requested):
One small thing, otherwise LGTM thanks :+1: 

Looks like the CI failure is from calloop, which uses thiserror as well and apparently does not specify its minimal version correctly.

wayland-scanner/src/lib.rs
bjorn3: Moved

bjorn3:
Rebased

PR #587, opened 2022-12-09
Build documentation with all features for docs.rs

hw0lff:
Not all crates were build with all features enabled for docs.rs. So some of the hyperlinks broke.
This adds
```toml
[package.metadata.docs.rs]
all-features = true
rustdoc-args = ["--cfg", "docsrs"]
```
to the Cargo.toml of all wayland-rs crates except the `wayland-tests` crate.


vberger (approved):
Oh right, thank you!

PR #589, opened 2022-12-16
server: Implement `AsFd` for `ListeningSocket`

ids1024:
It looks like the only other `AsRawFd` implementations here are in wayland-backend and already have corresponding `AsFd` implementations.


vberger:
Thanks!

PR #594, opened 2023-01-01
wayland-client: fix compilation without `log`

kchibisov:
This should use `eprintln!` when the log
feature is not present with `calloop`.


kchibisov:
The build failure is irrelevant.

PR #595, opened 2023-01-04
protocols: add gtk-shell1

technobaboo:


vberger:
Thanks, could you just add a changelog entry for this?

wayland-protocols-misc/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+## 0.2.0 -- 04/01/2023
```
vberger, line 5: No, sorry if that was unclear, don't add a new header, I do that when doing the release. :)

PR #598, opened 2023-01-22
server: Add resource and display handle to Dispatch::destroyed

i509VCB:
This allows libraries (such as smithay) which may dispatch events to users when a protocol object is to destroyed to eliminate lists of known objects to destroy.

This is an example of some of the code in Smithay that can be removed with these changes

`src/wayland/shell/xdg/handlers/surface/toplevel.rs`

```diff
-    fn destroyed(state: &mut D, _client_id: ClientId, object_id: ObjectId, data: &XdgShellSurfaceUserData) {
+    fn destroyed(state: &mut D, _client_id: ClientId, toplevel: &XdgToplevel, data: &XdgShellSurfaceUserData) {
        data.alive_tracker.destroy_notify();
        data.decoration.lock().unwrap().take();

-        let mut shell_data = data.shell_data.lock().unwrap();
-        if let Some(index) = shell_data
-            .known_toplevels
-            .iter()
-            .position(|top| top.shell_surface.id() == object_id)
-        {
-            let toplevel = shell_data.known_toplevels.remove(index);
-            drop(shell_data);
-            let surface = toplevel.wl_surface().clone();
            XdgShellHandler::toplevel_destroyed(state, toplevel);
            compositor::with_states(&surface, |states| {
                *states
                    .data_map
                    .get::<XdgToplevelSurfaceData>()
                    .unwrap()
                    .lock()
                    .unwrap() = Default::default();
                *states.cached_state.pending::<SurfaceCachedState>() = Default::default();
                *states.cached_state.current::<SurfaceCachedState>() = Default::default();
            })
-        }
    }
```

### DIFF
--- a/wayland-backend/src/rs/server_impl/handle.rs
+++ b/wayland-backend/src/rs/server_impl/handle.rs
@@ -41,14 +41,19 @@ impl<D> State<D> {
         }
     }
 
-    pub(crate) fn cleanup(&mut self) -> impl FnOnce(&mut D) {
+    pub(crate) fn cleanup<'a>(&mut self, handle: &'a super::Handle) -> impl FnOnce(&mut D) + 'a {
         let dead_clients = self.clients.cleanup(&mut self.pending_destructors);
         self.registry.cleanup(&dead_clients);
         // return a closure that will do the cleanup once invoked
         let pending_destructors = std::mem::take(&mut self.pending_destructors);
         move |data| {
             for (object_data, client_id, object_id) in pending_destructors {
-                object_data.destroyed(data, ClientId { id: client_id }, ObjectId { id: object_id });
+                object_data.clone().destroyed(
+                    handle,
+                    data,
+                    ClientId { id: client_id },
+                    ObjectId { id: object_id },
+                );
             }
         }
     }
--- a/wayland-server/src/dispatch.rs
+++ b/wayland-server/src/dispatch.rs
@@ -111,7 +111,13 @@ pub trait Dispatch<I: Resource, UserData, State = Self>: Sized {
     /// convenience.
     ///
     /// By default this method does nothing.
-    fn destroyed(_state: &mut State, _client: ClientId, _resource: ObjectId, _data: &UserData) {}
+    fn destroyed(
+        _state: &mut State,
+        _client: wayland_backend::server::ClientId,
+        _resource: &I,
+        _data: &UserData,
+    ) {
+    }
 }
 
 /// The [`ObjectData`] implementation that is internally used by this crate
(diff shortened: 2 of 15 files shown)


wayland-server/src/dispatch.rs
```diff
@@ -252,11 +259,15 @@ impl<I: Resource + 'static, U: Send + Sync + 'static, D: Dispatch<I, U> + 'stati
...
     fn destroyed(
         &self,
+        handle: &wayland_backend::server::Handle,
         data: &mut D,
         client_id: wayland_backend::server::ClientId,
         object_id: wayland_backend::server::ObjectId,
     ) {
-        <D as Dispatch<I, U>>::destroyed(data, client_id, object_id, &self.udata)
+        let dhandle = DisplayHandle::from(handle.clone());
+        let resource = I::from_id(&dhandle, object_id).unwrap();
```
vberger, line 268: Given we are invoking it on a dead object, this will produce and inert handle, from which the user data cannot be accessed. Is this what we want here?

i509VCB: Probably not that specific behavior.

I'd however argue that wayland-backend changing it's behavior around object destruction to allow the user data to be accessible would be useful.

At the moment we have a few handlers in smithay which require tracking every protocol object that has been seen on the past. https://github.com/Smithay/smithay/blob/master/src/wayland/shell/xdg/handlers/surface/toplevel.rs#L125 since object IDs are still equal even after destruction, this can work to recover the object.

However even if the workaround suffices, this type of pattern is going to be needed for every protocol object type if we want to consider eliminating IsAlive in smithay or heavily reducing it's presence.

vberger: To do that properly, we'd need to change the API of the `ObjectData` trait so that `destroyed()` takes an `Arc<Self>` instead of a `&self`, and use the doc-hidden `__set_object_data` method of the newly created resource to insert the `self` in it.

That'll be breaking change for both wayland-backend and wayland-server, though I'm thinking if we're doing that server side, we should also do it client side.

i509VCB: I've started working on that. I do agree the client should have the same changes but I will do the client in a later pull request.

wayland-backend/src/rs/server_impl/handle.rs
```diff
@@ -41,14 +41,19 @@ impl<D> State<D> {
...
         let pending_destructors = std::mem::take(&mut self.pending_destructors);
         move |data| {
             for (object_data, client_id, object_id) in pending_destructors {
-                object_data.destroyed(data, ClientId { id: client_id }, ObjectId { id: object_id });
+                object_data.clone().destroyed(
+                    handle,
+                    data,
+                    ClientId { id: client_id },
+                    ObjectId { id: object_id },
+                );
```
elinorbgr, line 56: Actually, I think it would be better to instead change the returned closure type to take both `data` and `handle` as argument, rather than capture the handle in the closure. Would make the API less confusing wrt what borrows what or which locks are held when what is invoked.

elinorbgr:
I rebased and made the final edits, closing in favor of #654

PR #599, opened 2023-01-22
server: allow posting errors in GlobalDispatch on global object

i509VCB:


wayland-server/src/global.rs
```diff
@@ -29,26 +29,32 @@ impl<I: Resource + 'static, U: Send + Sync + 'static, D: GlobalDispatch<I, U> +
...
         );
 
         match new_data {
-            Some(data) => data,
-            None => panic!(
+            Some(data) => Some(data),
+            None if protocol_error.is_some() => {
+                let error = protocol_error.unwrap();
+                resource.post_error(error.0, error.1);
+                None
```
vberger, line 55: Couldn't we just have this function set some dummy object_data (invoking `custom_init` internally), rather than adding options all over the API ?

Also, this match can be made cleaner by matching over `(new_data, protocol_error)` rather than using an if and an unwrap.

vberger (approved):

PR #607, opened 2023-03-17
scanner: Implement `Hash`

ids1024:
https://github.com/Smithay/wayland-rs/pull/603 implemented `Borrow`, but that still doesn't allow directly using generated types as hashmap keys. Given `ObjectId` already implements `Hash`, and the wrappers already implement `Eq`, this easy easy enough, and shouldn't have any issues those other trait implementations don't also have.


vberger:
Looks good, thanks! Could you just add a changelog entry for that change?

vberger:
Thanks!

PR #608, opened 2023-03-18
Update wayland.xml to 1.21.92

chrisduerr:
I think the most noteworthy change here is probably the addition of `preferred_scale`, which I think makes some sense to integrate considering fractional scale is already supported and implementing preferred scale on top of that should be pretty simple for compositors.

### DIFF
--- a/wayland-client/CHANGELOG.md
+++ b/wayland-client/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Updated Wayland core protocol to 1.21.92
+
 ## 0.30.1 -- 04/02/2023
 
 #### Bugfixes
--- a/wayland-client/CHANGELOG.md
+++ b/wayland-client/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


vberger:
1.21.92 is a beta release, shouldn't we wait for the proper 1.22 release?

In any case, could you add changelog entries to wayland-client and wayland-server covering this update?

chrisduerr:
> 1.21.92 is a beta release, shouldn't we wait for the proper 1.22 release?

That's something I was pondering myself, I'm not sure about your policy on this.

To me it seems unlikely that there will be major changes to protocols after they've been officially released, but I'm not familiar with Wayland's track record here.

But since new additions don't introduce breaking changes to existing protocols, I think it should be fine to expose unstable additions like this. Wayland protocol extensions are already heavily used in unstable revisions and I think since most compositor/client developers directly using this are likely looking at the protocols themselves, they can make their own decisions about relying on unstable/beta versions.

elinorbgr:
Ok, so 1.22 has been released now, with no changes since 1.21.92. I'm good with merging this, maybe let's just adjust the changelogs accordingly?

elinorbgr:
Thanks

PR #610, opened 2023-03-22
Update Rust MSRV to 1.64.0

ids1024:
This should fix the CI failure due to raw-window-handle requiring Rust 1.64.0.


vberger:
I'm not sure about this one, we don't really need to actually bump the MSRV just for this feature. I'd rather change the CI script to ignore that feature when checking the MSRV.

elinorbgr:
ci was fixed in #617

PR #612, opened 2023-04-02
Implement TODO

WhyNotHugo:
Came across this TODO trying to understand the logic in here and figured it was a pretty trivial fix.

### DIFF
--- a/wayland-client/src/globals.rs
+++ b/wayland-client/src/globals.rs
@@ -155,15 +155,9 @@ impl GlobalList {
         let (name, version) = guard
             .iter()
             // Find the with the correct interface
-            .filter_map(|Global { name, interface: interface_name, version }| {
-                // TODO: then_some
-                if interface.name == &interface_name[..] {
-                    Some((*name, *version))
-                } else {
-                    None
-                }
+            .find_map(|Global { name, interface: interface_name, version }| {
+                (interface.name == interface_name.as_str()).then_some((*name, *version))
             })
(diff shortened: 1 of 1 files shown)


elinorbgr:
Unfortunately, this change would bump the MSRV from 1.59 to 1.62, and I don't think this is by itself a sufficient reason to do it. :sweat_smile:

WhyNotHugo:
Odd, the error message indicates that this is due to the `raw-window-handle` crate:

> error: package `raw-window-handle v0.5.2` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.59.0

But yeah, [`then_some`](https://doc.rust-lang.org/std/primitive.bool.html#method.then_some) requires 1.62.

I'll just leave this as draft until the MSRV is pushed for some other reason.

WhyNotHugo:
This might actually happen soon due to https://github.com/Smithay/wayland-rs/pull/610

elinorbgr:
> error: package raw-window-handle v0.5.2 cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.59.0

Yeah, as said on the other PR you linked this is actually a CI issue, I don't plan to bump the MSRV because of an optional dep.

elinorbgr:
Closing this following #617

WhyNotHugo:
I left this in draft because it relied on a newer MSRV. Is the change itself entirely unwanted?

PR #619, opened 2023-04-27
feat: add a layershellexample

Decodetalkers:
I found that layershell is very similar with xdg, so I want to add one example of layershell with origin wayland-rs..because the are so similar

### DIFF
--- a/wayland-client/examples/simple_layershell.rs
+++ b/wayland-client/examples/simple_layershell.rs
@@ -1,0 +1,231 @@
+use std::{fs::File, os::unix::io::AsFd};
+use wayland_client::{
+    delegate_noop,
+    protocol::{
+        wl_buffer, wl_compositor, wl_keyboard, wl_registry, wl_seat, wl_shm, wl_shm_pool,
+        wl_surface,
+    },
+    Connection, Dispatch, Proxy, QueueHandle, WEnum,
+};
+
+use wayland_protocols_wlr::layer_shell::v1::client::{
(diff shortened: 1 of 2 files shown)


wayland-client/examples/simple_layershell.rs
```diff
@@ -0,0 +1,289 @@
...
+        _: &mut Self,
+        _: &wl_compositor::WlCompositor,
+        _: wl_compositor::Event,
+        _: &(),
+        _: &Connection,
+        _: &QueueHandle<Self>,
+    ) {
+        // wl_compositor has no event
+    }
+}
```
WhyNotHugo, line 117: Want to use `delegate_noop!` here instead?

```suggestion
delegate_noop!(State: wl_compositor::WlCompositor);
```

Since this is an example, let's use the approach that's simplest for other to follow :)

Decodetalkers: but I want it be the same as xdg-shell, emm, to show they are similar

Eclextic:
I just wanted to say, thank you very much for this example! It has really helped me out and guided me! Why exactly has it not yet been merged?? This is gold!

wayland-client/examples/simple_layershell.rs
WhyNotHugo: Might be best to update that example to use `delegate_noop!` too.

Decodetalkers: ok, done

PR #624, opened 2023-05-18
Update wayland-protocols to 1.32

WhyNotHugo:
This relies on unreleased (but already merged) changes upstream. I'm not sure if you'd rather merge this now or wait for an upstream release, but I'm sharing this at this point since I want to write application code that uses this protocol.

### DIFF
--- a/wayland-protocols/src/ext.rs
+++ b/wayland-protocols/src/ext.rs
@@ -37,3 +37,24 @@ pub mod session_lock {
         );
     }
 }
+
+#[cfg(feature = "staging")]
+pub mod foreign_toplevel_list {
+    //! The purpose of this protocol is to provide protocol object handles for toplevels, possibly
+    //! originating from another client.
+    //!
+    //! This protocol is intentionally minimalistic and expects additional functionality (e.g.
+    //! creating a screencopy source from a toplevel handle, getting information about the state of
+    //! the toplevel) to be implemented in extension protocols.
+    //!
+    //! The compositor may choose to restrict this protocol to a special client launched by the
+    //! compositor itself or expose it to all clients, this is compositor policy.
+
+    #[allow(missing_docs)]
+    pub mod v1 {
+        wayland_protocol!(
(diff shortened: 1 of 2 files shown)


WhyNotHugo:
CI failures seem unrelated.

WhyNotHugo:
> In general I prefer to follow wayland-protocols upstream releases.

Of course, we can wait until there is a stable release before merging this.

I mostly created this branch to unblock myself so that I can start using this protocol on an application of mine (shotman), but figured I might avoid someone else duplicating the work by leaving this PR open here.

I don't know when an upstream release will happen.

kchibisov:
fyi, the protocol is already available in sctk branch, given that it was used as a reference for wayland-protocols in the first place https://github.com/Smithay/client-toolkit/pull/381

i509VCB:
There has been a wayland-protocols release recently if I recall, so we probably want to do a full update.

i509VCB:
If possible could you also add the other two protocols introduced in 1.32?

WhyNotHugo:
Since rebasing on `master`, none of the docs are generating properly for me.

Cursor shape is empty:

image

But so are other protocols like idle_notify:

image

WhyNotHugo:
Updated PR with all protocols from wayland-protocols 1.32.

kchibisov:
@i509VCB ping.

i509VCB (approved):
Apparently `log` raised MSRV so msrv build fails.

WhyNotHugo:
The cause of the CI failure is in https://github.com/Smithay/wayland-rs/blob/master/.gitignore#L3

Because there is no lockfile (and CI is also not instructed to use `--frozen`), the latest version of dependencies is pulled on each run.

The line I've linked should be removed and the lockfile should be committed to the repository to prevent this kind of breakage from happening again. To deal with this particular issue, it's either possible to bump the MSRV or pin an older `log` crate. I can send a PR for either alternative.

PR #626, opened 2023-05-21
Update memoffset to 0.9

est31:
Updates memoffset to version 0.9


elinorbgr:
Yeah, wayland-test tests should not be ran as part of the whole workspace tests, as it causes some inconsistency in the cargo features. That's why the CI runs the workspace tests with `--exclude wayland-test`, and then the wayland-test tests by themselves afterwards.

Issue #604, opened 2023-02-27
Remove lazy_static from the dependency tree


notgull:
The Rust ecosystem is moving away from `lazy_static` in favor of `OnceCell` et al. The `wayland-sys` crate is the only case of `lazy_static` in my dependency tree, so it should move towards `once_cell` or another crate.

The issue is that `lazy_static` is a public dependency of `wayland-sys`, so it can't be removed easily. The best way to avoid it would be to have the `WAYLAND_CLIENT_HANDLE` et al become:

```rust
pub fn wayland_client_handle() -> &'static WaylandClient {
    static WAYLAND_CLIENT_HANDLE: OnceCell<WaylandClient> = OnceCell::new();
    WAYLAND_CLIENT_HANDLE.get_or_init(|| /* ... */)
}
```

That way, future changes wouldn't be breaking.

PR #627, opened 2023-05-28
Replace `lazy_static` with `once_cell`

daxpedda:
This replaces [`lazy_static`](https://crates.io/crates/lazy_static) with [`once_cell`](https://crates.io/crates/once_cell). See #604 for more information.

Initially I was intending to not expose the `Lazy` type like we did before with types generated by `lazy_static`, but turned out that libraries like [`dlib`](https://crates.io/crates/dlib) don't play well with functions, so I left it as is. Happy to change it to a function though.

Because of the currently exposed `lazy_static` types, this is a breaking change.

Fixes https://github.com/Smithay/wayland-rs/issues/604.
Requires https://github.com/elinorbgr/dlib/pull/21.


daxpedda:
CI fails with the same errors as on master, unrelated to this PR.

notgull:
I'd also like not to expose `once_cell`, since it may be replaced by `std` in the near future. It would be nice if we could just rework `dlib` to support this use case.

daxpedda:
Made https://github.com/elinorbgr/dlib/pull/21 to address this.

elinorbgr:
The dlib PR is merged, and I released dlib 0.5.1 with that change.

daxpedda:
Hm, I guess we can just ignore the code coverage failure here?

elinorbgr:
Ah wait, could you just add an entry to wayland-sys's changelog? this is a breaking change after all.

daxpedda:
Done.

elinorbgr:
Perfect, thanks!

PR #628, opened 2023-06-05
Fix CI

daxpedda:
Apparently `pub extern crate` requires docs with the `missing_docs` lint and you can't use `#[allow(...)]` on it. So replacing it with `pub use` fixes this issue.

Also sprinkled some `#[allow(clippy::redundant_clone)]` until https://github.com/rust-lang/rust-clippy/pull/10873 reaches stable.

### DIFF
--- a/wayland-backend/src/lib.rs
+++ b/wayland-backend/src/lib.rs
@@ -47,8 +47,10 @@
 // Doc feature labels can be tested locally by running RUSTDOCFLAGS="--cfg=docsrs" cargo +nightly doc -p <crate>
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
-pub use io_lifetimes;
-pub use smallvec;
+/// Reexport of the `io_lifetimes` crate, which is part of `wayland-backend`'s public API.
+pub extern crate io_lifetimes;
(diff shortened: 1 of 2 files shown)


wayland-backend/src/lib.rs
```diff
@@ -47,10 +47,8 @@
 // Doc feature labels can be tested locally by running RUSTDOCFLAGS="--cfg=docsrs" cargo +nightly doc -p <crate>
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
-#[allow(missing_docs)]
-pub extern crate io_lifetimes;
-#[allow(missing_docs)]
-pub extern crate smallvec;
+pub use io_lifetimes;
+pub use smallvec;
```
elinorbgr, line 51:
```suggestion
/// Reexport of the `io_lifetimes` crate, which is part of `wayland-backend`'s public API.
pub extern crate io_lifetimes;
/// Reexport of the `smallvec` crate, which is part of `wayland-backend`'s public API.
pub use smallvec;
```

elinorbgr:
Actually it looks like even nightly rustdoc ignores documentation of `pub extern crate` items when generating the documentation. Not sure why this lint was enabled then. :woman_shrugging:

elinorbgr:
Yes, thanks!

PR #629, opened 2023-06-05
Bump `bitflags` to v2

### LINKED ISSUE #622: Update bitflags to 2.0
This will be a breaking change, because they are a part of public interface(expect for wayland-backend). Probably better to include as a part of other potential breaking changes.

daxpedda:
Changelog: https://github.com/bitflags/bitflags/blob/2.0.0/CHANGELOG.md#200.

This is a breaking change. The API that is implemented by the macro is now different.

We didn't really have to derive anything apart from `Debug` to make it work, but considering this is public I derived all other traits as well except `PartialOrd`, `Ord` and `Default`.

Fixes #622.

### DIFF
--- a/wayland-client/CHANGELOG.md
+++ b/wayland-client/CHANGELOG.md
@@ -1,8 +1,12 @@
 # CHANGELOG: wayland-client
 
-## 0.30.2 -- 30/05/2023
-
 ## Unreleased
+
+#### Breaking changes
+
+- Bump bitflags to 2.0
+
+## 0.30.2 -- 30/05/2023
 
 - Updated Wayland core protocol to 1.22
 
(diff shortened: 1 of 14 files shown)


daxpedda:
CI failure is actually an issue in `calloop` with minimal dependencies: https://github.com/Smithay/calloop/pull/133.

daxpedda:
What should I do here? Are older versions of `calloop` going to be yanked or should I just raise the minimum version of `calloop`?

elinorbgr:
For this PR, just bump the calloop dependency to `0.10.6` (I released both 0.11 and 0.10.6, and both have the fix).

regarding calloop and wayland-client in general, I need to check back *why* we included them in the first place, and this is out of scope for this PR.

daxpedda:
Done.

kchibisov:
>regarding calloop and wayland-client in general, I need to check back why we included them in the first place, and this is out of scope for this PR.

It was included because it was 1) common code which could be used by other libs outside of the client-toolkit, so the code from client toolkit moved up. 2) Yes, it's for other PR, but in general, I think it should be separate 1 file crate with just a source.

PR #632, opened 2023-06-22
server: Verify that .lock fd matches on-disk file

YaLTeR:
Fixes a race condition that can happen if one thread unlinks the .lock between our open() and flock() calls.


elinorbgr:
Sorry for the late review. Thanks, this looks good. Could you just add a changelog entry for this?

PR #635, opened 2023-07-08
server: Add `flush` to `DisplayHandle`, and implement `AsFd` for `Display`, for use in smithay

ids1024:
This makes it easy to use `wayland_server::Display` with `calloop` 0.11 without hacking around IO safety.

If these solutions don't seem ideal for any reason, there may be alternatives. But some change is needed.


elinorbgr:
Thanks, looks good :+1: 

could you add a changelog entry as well for those changes?

PR #636, opened 2023-07-12
Prefer sonames when loading dynamic libraries

fredizzimo:
The standard practice for loading dynamic libraries is to use `sonames`, versioned `.so` files. That way you ensure that the `.so` file contains the expected API, and not another version with possibly breaking changes. Therefore, the soname version should be tried first, and only then as a backup the unversioned `.so` file.

Doing so is not a workaround for Ubuntu, like the comments suggest, virtually all Linux distributions use sonames, so those comments have also been removed.

For me this order caused problems when making a Snap package for Neovide, using the classic confinement. On Ubuntu and also in Snaps, the unversioned `.so` file is only installed if you install the developer packages, so the Snap does not include those. Therefore the `.so` file is not found and loaded from the host instead, despite the fact that the correct file is included in the Snap. Note that a similar change also had to be done for https://github.com/rust-windowing/xkbcommon-dl/pull/16.

It would be great if a patch release could be released with these changes, so that I can continue working on the Neovide Snap properly.


elinorbgr:
Looks good, thanks. I'll make a quick release with this change. :+1:

Issue #592, opened 2022-12-20
Use `BorrowedFd` for requests/events instead of `RawFd`


ids1024:
Currently events a client receives with an fd use `OwnedFd`, while events the client sends use `RawFd`. In server code it's the same but reversed.

As a result of this, functions like `WlShm::create_pool` violate IO safety rules since the function takes and operates on a `RawFd`, but the function is not marked `unsafe`.

This should instead use a `BorrowedFd`, but this is complicated from a lifetime perspective. I think `wayland_client::Proxy`'s `Request` and `wayland_server::Resource`'s `Event` would need to use GATs generic over a lifetime? This isn't possible as long as wayland-rs targets Rust < 1.65.0 though.

vberger:
Yes, this lifetime constraint is the exact reason why I left the `RawFd` at this place.

ids1024:
I guess this is something for 0.31.0, by which point wayland-rs's Rust MSRV will probably be new enough to use GATs.

elinorbgr:
This is now done.

PR #640, opened 2023-07-13
Use `BorrowedFd` in arguments

ids1024:
Implementation of https://github.com/Smithay/wayland-rs/issues/592.

This is required for IO safety. Or perhaps `AsFd` generics would be useful.

This will be a breaking API change, and require Rust 1.65.

Currently `wayland-client` compiles here, but server-side needs the same changes, and things need to be cleaned up a bit.


ids1024:
Hm, an awkward API stability concern here: adding `Request`/`Event` enums are non-exhastive. So adding a variant is considered a non-breaking change. But as this is currently implemented in this PR, adding a variant with an FD adds a type generic, which is breaking.

That is needed because an enum can't have a type generic unless a variant uses that type parameter. Maybe if it's non-exhaustive anyway, it could have a variant that's never constructed with a `PhantomData`? And an uninhabited type for good measure.

ids1024:
Currently `Request`/`Event` types and methods also accept `String` and `Vec<u8>` instead of `&str` and `&[u8]`. Not that much of an issue since those types aren't the most common in Wayland, but presumably with a GAT here those could also be changed, avoiding an extra allocation and provide a more idiomatic API.

ids1024:
Added generation of a `#doc(hidden)` phantom variant for API stability.

This should be good now.

elinorbgr (approved):
That's a solid PR, thanks!

PR #642, opened 2023-07-22
Replace io-lifetimes with std::os::fd

linkmauve:
Since Rust 1.63, most of the features of io-lifetimes have been merged into std, including the only three we are using: `OwnedFd`, `BorrowedFd` and the `AsFd` trait.

This bumps our MSRV, but this is the MSRV of the newest io-lifetimes anyway.

I kept reexporting `io_lifetimes` as it was part of the public interface, but since it isn’t used any more it would be nice to remove it before the next minor version bump.  I can prepare a commit for that.

This partially conflicts with #640.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The documentation for the releases can be found on [docs.rs](https://docs.rs/):
 
 ## Requirements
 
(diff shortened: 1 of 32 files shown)


elinorbgr (approved):
Sorry for the late review. Thanks, I'll merge this and build upon it. :+1:

PR #650, opened 2023-08-25
backend: fix prepare_read behavior

elinorbgr:
This should fix the multithreading issues that have been encountered.

cc @kchibisov

### DIFF
--- a/wayland-client/src/conn.rs
+++ b/wayland-client/src/conn.rs
@@ -136,7 +136,8 @@ impl Connection {
     ///
     /// If you don't need to manage multiple event sources, see
     /// [`blocking_dispatch()`](EventQueue::blocking_dispatch) for a simpler mechanism.
-    pub fn prepare_read(&self) -> Result<ReadEventsGuard, WaylandError> {
+    #[must_use]
+    pub fn prepare_read(&self) -> Option<ReadEventsGuard> {
         self.backend.prepare_read()
     }
 
@@ -162,15 +163,11 @@ impl Connection {
         loop {
             self.backend.flush()?;
 
-            // first, prepare the read
-            let guard = self.backend.prepare_read()?;
-
-            // If another thread processed events prior to prepare_read, we might already be done.
-            if done.done.load(Ordering::Relaxed) {
-                break;
+            if let Some(guard) = self.backend.prepare_read() {
+                dispatched += blocking_read(guard)?;
+            } else {
+                dispatched += self.backend.dispatch_inner_queue()?;
             }
-
-            dispatched += blocking_read(guard)?;
 
             // see if the successful read included our callback
             if done.done.load(Ordering::Relaxed) {
--- a/wayland-backend/CHANGELOG.md
+++ b/wayland-backend/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - MSRV bumped to 1.65
 - `io-lifetimes` is no longer a (public) dependency
+- The `Backend::prepare_read()` method now returns `None` if the inner queue of the backend
+  needs to be dispatched using `Backend::dispatch_inner_queue()`, instead of trying to dispatch
+  it by itself. This can only happen when using the `sys` backend, and allows the crate to
+  behave properly when multiple threads try to read the socket using the libwayland API.
 
 #### Additions
 
--- a/wayland-backend/src/client_api.rs
+++ b/wayland-backend/src/client_api.rs
@@ -251,27 +251,24 @@ impl Backend {
     /// This is the first step for actually reading events from the Wayland socket. See
     /// [`ReadEventsGuard`] for how to use it.
     ///
-    /// This call will not block, but event callbacks may be invoked in the process
-    /// of preparing the guard.
-    ///
-    /// If it returns an error, you Wayland connection is already dead.
+    /// This call will not block, but may return `None` if the inner queue of the backend needs to
+    /// be dispatched. In which case you should invoke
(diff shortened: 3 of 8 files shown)


wayland-client/src/conn.rs
```diff
@@ -136,7 +136,7 @@ impl Connection {
     ///
     /// If you don't need to manage multiple event sources, see
     /// [`blocking_dispatch()`](EventQueue::blocking_dispatch) for a simpler mechanism.
-    pub fn prepare_read(&self) -> Result<ReadEventsGuard, WaylandError> {
+    pub fn prepare_read(&self) -> Option<ReadEventsGuard> {
```
kchibisov, line 139: Should it be marked as `#[must_use]`?

wayland-client/src/conn.rs
```diff
@@ -162,16 +162,12 @@ impl Connection {
...
-            // first, prepare the read
-            let guard = self.backend.prepare_read()?;
-
-            // If another thread processed events prior to prepare_read, we might already be done.
-            if done.done.load(Ordering::Relaxed) {
-                break;
+            if let Some(guard) = self.backend.prepare_read() {
+                dispatched += blocking_read(guard)?;
+            } else {
+                dispatched += self.backend.dispatch_inner_queue()?;
```
kchibisov, line 168: Hm, but you could have a blocking dispatch of the queue here as well? Non-blocking roundtrip doesn't sound like it'll work reliably or I don't understand something?

kchibisov (changes requested):
Probably the same about `#[must_use]` for other methods. Also, maybe we should use `Result` instead, but use some `io` type to indicate that lock can't be acquired? The mutex stuff uses `Result` and it's more intuitive than the `Option`.

wayland-client/src/conn.rs
elinorbgr: good point :+1:

wayland-protocols/CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 ### Breaking changes
 
 - Bump bitflags to 2.0
-- Updated wayland-backend to 0.2
+- Updated wayland-backend to 0.3
```
kchibisov, line 8: This looks arbitrary and unrelated to this PR, should be submitted separately.

wayland-client/src/conn.rs
elinorbgr: This is in a loop, so it'll just try again at the next iteration.

wayland-protocols/CHANGELOG.md
elinorbgr: Well, this PR introduces breaking changes that cascade to all other crates, so it seems appropriate to add the associated changelog entries in the same PR.

wayland-client/src/conn.rs
kchibisov: Ah, yeah, sure, I just know that there's mo efficient API to do the exact same thing, but I think it's not an option for us. The `wl_display_dispatch_queue` https://wayland.freedesktop.org/docs/html/apb.html#Client-classwl__display_1ae027b09801474ac7c6b0f1ef25ff6e17

wayland-protocols/CHANGELOG.md
kchibisov: But you don't actually bump the version here, so it should be done in a separate commit.

wayland-client/src/conn.rs
elinorbgr: That function in wayland-client does mostly exactly the same thing as `wl_display_roundtrip`, but implemented in terms of wayland-backend's API, so I don't really get your point?

wayland-protocols/CHANGELOG.md
elinorbgr: ok

elinorbgr:
> Also, maybe we should use Result instead, but use some io type to indicate that lock can't be acquired?

Not sure what we would gain from this? It feels to me that this `Option<_>` reflects best the behavior of this method, changing it to a `Result` we would just have an error type identical to `()`, what would be the point then?

wayland-client/src/conn.rs
kchibisov: `wl_display_roundtrip` uses `poll` under the hood to not do the loop.

kchibisov (approved):

wayland-client/src/conn.rs
elinorbgr: What... ?

`wl_display_roundtrip` does have a loop: https://gitlab.freedesktop.org/wayland/wayland/-/blob/main/src/wayland-client.c#L1402

and `blocking_read` uses `poll` internally: https://github.com/Smithay/wayland-rs/blob/master/wayland-client/src/conn.rs#L224-L245

kchibisov: @elinorbgr if you look closely what their code is and what we have, they use blocking dispatch. However we use it only in `read_guard` branch, and the other branch is using regular non-blocking dispatch.

So while they have a `loop`, the `wait`, but we `busy loop` from the `else`. The blocking read is fine, the `else` is a busy loop. `dispatch_inner_queue` is a non_blocking API from what I can, see, but libwayland always using blocking API, with `dispatch_queue`.

kchibisov: After reading the libwayland impl it's pretty much the same, when their docs are not really clear what exactly blocks and how it works.

PR #659, opened 2023-09-23
server: Use `std::fs` APIs instead of `nix` where possible

ids1024:


wayland-server/src/socket.rs
```diff
@@ -83,18 +82,16 @@ impl ListeningSocket {
...
-
-            // SAFETY: We have just opened the file descriptor.
-            _lock = unsafe { File::from_raw_fd(lock_fd) };
+            _lock = File::options()
+                .create(true)
+                .read(true)
+                .write(true)
+                .mode(0o660)
+                .open(&lock_path)
+                .map_err(|_| BindError::PermissionDenied)?;
```
elinorbgr, line 91: `CLOEXEC` set by default by the `File` constructor I assume?

ids1024: Yes, it should set `CLOEXEC` regardless of what options are used.

PR #660, opened 2023-09-23
WIP/experiment: use as-any instead of downcast-rs

fogti:
(this is to compare feature parity between `as-any` and `downcast-rs`, the only thing that is afaik currently missing is `into_any_arc`)


ids1024:
With trait upcasting in Rust 1.86, we *almost* don't need downcast-rs, so https://github.com/Smithay/wayland-rs/pull/912 removed it and defined an `AsAny` trait to workaround the one single place this is still a problem.

PR #666, opened 2023-10-04
Use `VecDeque<OwnedFd>` and `Vec<OwnedFd>` for `in_fds`/`out_fds`

ids1024:
This avoids some awkward and potentially error-prone code, and should perform no worse than how the `Buffer` type was used.

### DIFF
--- a/wayland-backend/src/rs/socket.rs
+++ b/wayland-backend/src/rs/socket.rs
@@ -1,7 +1,8 @@
 //! Wayland socket manipulation
 
+use std::collections::VecDeque;
 use std::io::{ErrorKind, IoSlice, IoSliceMut, Result as IoResult};
-use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, IntoRawFd, OwnedFd, RawFd};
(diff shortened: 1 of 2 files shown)


ids1024:
https://github.com/Smithay/wayland-rs/pull/672 and https://github.com/Smithay/wayland-rs/pull/673 address CI failures here, that aren't related to this PR.

This should be good to merge now. A nice reduction in code, if not by that much. Despite this being PR 666, this seems less cursed than the code that was using `std::mem::forget`.

ids1024:
Hm, I guess `parse_message` here could behave incorrectly if `MssingData` is returned after consuming an fd. Then the fd will have been discarded when it tries to parse again later.

Moving to `OwnedFd` where possible is good, though that is an awkward case to deal with...

ids1024:
Okay, it already checks `len > raw.len()` before parsing, so that would only happen if a message contains two fds and all the bytes are written before the second fd is sent. And this issue already exists. So I guess that doesn't block this.

elinorbgr:
I guess this needs to be rebased for the CI to properly run

PR #671, opened 2023-10-29
Include advertised version in `BindError::UnsupportVersion`

hw0lff:
Reporting the advertised version is helpful because I can then report this to the user in case of an error.
This way I can get more information in a reported issue.

I also added two code clean up commits since the code was right next to each other.

The tests did not run successfully but they don't run through on master either.
Both branches fail at `client_receive_generic_error`:
```
---- client_receive_generic_error stdout ----
Protocol error 42 on object wl_compositor@3:
Protocol error 42 on object wl_compositor@3:
thread 'client_receive_generic_error' panicked at wayland-tests/tests/protocol_errors.rs:56:9:
assertion `left == right` failed
  left: ""
 right: "I don't like you!"
```

### DIFF
--- a/wayland-client/src/globals.rs
+++ b/wayland-client/src/globals.rs
@@ -168,7 +168,7 @@ impl GlobalList {
 
         // Test version requirements
         if version < version_start {
-            return Err(BindError::UnsupportedVersion);
+            return Err(BindError::UnsupportedVersion(version));
         }
 
         // To get the version to bind, take the lower of the version advertised by the server and the maximum
@@ -232,7 +232,7 @@ impl From<InvalidId> for GlobalError {
 #[derive(Debug)]
 pub enum BindError {
     /// The requested version of the global is not supported.
-    UnsupportedVersion,
+    UnsupportedVersion(u32),
 
     /// The requested global was not found in the registry.
     NotPresent,
@@ -243,8 +243,8 @@ impl std::error::Error for BindError {}
 impl fmt::Display for BindError {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match self {
-            BindError::UnsupportedVersion {} => {
-                write!(f, "the requested version of the global is not supported")
+            BindError::UnsupportedVersion(version) => {
+                write!(f, "the available version {version} of the global is lower than the requested version")
             }
             BindError::NotPresent {} => {
                 write!(f, "the requested global was not found in the registry")
--- a/wayland-client/src/globals.rs
+++ b/wayland-client/src/globals.rs
@@ -156,12 +156,7 @@ impl GlobalList {
             .iter()
             // Find the with the correct interface
             .find_map(|Global { name, interface: interface_name, version }| {
-                // TODO: then_some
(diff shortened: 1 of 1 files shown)


wayland-client/src/globals.rs
```diff
@@ -243,8 +237,8 @@ impl std::error::Error for BindError {}
 impl fmt::Display for BindError {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match self {
-            BindError::UnsupportedVersion {} => {
-                write!(f, "the requested version of the global is not supported")
+            BindError::UnsupportedVersion(version) => {
+                write!(f, "the requested version {version} of the global is not supported")
```
elinorbgr, line 241: This wording suggests that `{version}` is the version that was requested by the user, not that advertised by the compositor.

elinorbgr:
The `client_receive_generic_error` is unrelated and the tests actually pass, it's just that cargo features interact badly with running tests for the whole workspace at once. The reference is the CI (on which the test pass as needed). I'm confused about the `check-minimal` CI job, but that seems unrelated to your PR.

wayland-client/src/globals.rs
hw0lff: Ah my mistake, I got confused there!
Should I remove the `{version}` completely or reorder the sentence to announce the actually available version?

elinorbgr: Maybe change the sentence so that it uses `version` but in a meaningful way

hw0lff: What do you think of `the available version {version} of the global is lower than the requested version`?

elinorbgr: sound good yes :+1:

wayland-client/src/globals.rs
```diff
@@ -232,7 +226,7 @@ impl From<InvalidId> for GlobalError {
 #[derive(Debug)]
 pub enum BindError {
     /// The requested version of the global is not supported.
-    UnsupportedVersion,
+    UnsupportedVersion(u32),
```
elinorbgr, line 229: This is a braking change, so please include a changelog entry for this

MarijnS95:
Would a similar change be accepted for `NotPresent`?

In this specific case I used to get:

```
thread 'main' panicked at .cargo/registry/src/index.crates.io-6f17d22bba15001f/smithay-client-toolkit-0.16.1/src/environment.rs:193:21:
[SCTK] A missing global was required: wl_subcompositor
```

On `winit 0.28` and all the Wayland dependencies it used back then, but nowdays on `winit 0.29` it is a way less descriptive:

```
called `Result::unwrap()` on an `Err` value: Os(OsError { line: 95, file: ".cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.29.4/src/platform_impl/linux/wayland/event_loop/mod.rs", error: WaylandError(Bind(NotPresent)) })
```

I'd like to include `(String)` with the `interface` name, if that's fine :)

ids1024:
Now that we're preparing for a major version release, this change has been merge in https://github.com/Smithay/wayland-rs/pull/682.

PR #677, opened 2023-11-20
backend: Use Buffer<u8> instead of Buffer<u32>

ids1024:
I don't think `sendmsg`/`recvmsg` are necessarily guaranteed to read/write a multiple of 4 bytes, so for correctness we need to buffer bytes rather than words. This also avoids unsafe casts.

This is based on https://github.com/Smithay/wayland-rs/pull/666. Not sure if the helpers here for reading/writing words to the buffer could be improved. Perhaps as methods of `Buffer`, if that's only being used for this.

This doesn't seem particularly worse to handle than the previous way, since either version has to deal with lengths in bytes vs. words, and array padding.


elinorbgr:
That looks pretty clean overall, is there something you want to change, given you've set this PR as a draft?

PR #678, opened 2023-11-20
backend: Use `Message<_, BorrowedFd>` rather than `RawFd`

ids1024:
This is a breaking change to `wayland-backend`, but not to `wayland-client` or `wayland-server`. I believe that should be fine.

This avoids an unsafe conversion in the `rs` backend.

Based on https://github.com/Smithay/wayland-rs/pull/677.

### DIFF
--- a/wayland-backend/src/client_api.rs
+++ b/wayland-backend/src/client_api.rs
@@ -2,7 +2,7 @@ use std::{
     any::Any,
     fmt,
     os::unix::{
-        io::{BorrowedFd, OwnedFd, RawFd},
+        io::{BorrowedFd, OwnedFd},
         net::UnixStream,
     },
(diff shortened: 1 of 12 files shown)


elinorbgr:
Make a breaking change in wayland-backend will force a following breaking release in wayland-client and wayland-server anyway, but I think this is a direction we want to go.

ids1024:
Oh right, those crates expose the `Backend` type, so it's a public dependency. Otherwise such a change would be non-breaking.

This change can be left until the next time wayland-rs has a breaking change. This isn't urgent since it isn't actually a soundness issue and shouldn't impact behavior. (And indeed, the fact this change works confirms that the `RawFd`s are already being used correctly.)

PR #679, opened 2023-11-29
feat: let delegate_noop can handle constGeneric

Decodetalkers:
the macro of delegate_noop cannot handle like
`struct A<const X: unsize>`, such kind of struct, so I want to fix it

### DIFF
--- a/wayland-client/CHANGELOG.md
+++ b/wayland-client/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 - Implement `AsFd` for `Connection` and `EventQueue` so they can easily be used in a
   `calloop` source.
+- Make `delegate_noop` usable for constGeneric type.
 
 ## 0.31.0 -- 2023-09-02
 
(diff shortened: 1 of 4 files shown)


Decodetalkers:
> Thanks, the principle looks good, I'll take some time later to carefully review the macro.
> 
> More generally for this change, would you mind:
> 
>     * Add some text in the documentation of the macro explaining how to use it
> 
>     * Introduce the same change in `wayland-server` as well which has a similar macro
> 
>     * Add relevant entries to the changelog of the two crates

Sure, I am happy to do this

elinorbgr:
I'm not sure about that, it seems to me that the pattern you used, `$(@< $( const $lt:tt $( : $clt:tt $(+ $dlt:tt )* )? ),+ >)?` would match a list like `const U: usize, const V: usize, ...`, but not a list mixing `const` arguments and regular type arguments?

Decodetalkers:
Ah.. I see, I fogot that generic types can be muti types..

Decodetalkers:
Ok, now it is ok.. I try add muti generic type, 

like 
```rust
struct A<const X: usize, Y: Sized, Z> {
     it: Z
}
```
it seems ok

PR #680, opened 2023-12-07
feat: load cursor from data

Decodetalkers:
I want to add a way to load cursor from raw data

### DIFF
--- a/wayland-cursor/src/lib.rs
+++ b/wayland-cursor/src/lib.rs
@@ -86,7 +86,9 @@ pub struct CursorTheme {
     fallback: Option<FallBack>,
 }
 
-struct FallBack(Box<dyn Fn(&str, u32) -> Option<Cow<'static, [u8]>>>);
+type FallBackInner = Box<dyn Fn(&str, u32) -> Option<Cow<'static, [u8]>>>;
+
+struct FallBack(FallBackInner);
 
 impl FallBack {
     fn new<F>(fallback: F) -> Self
@@ -195,9 +197,7 @@ impl CursorTheme {
             None => {
                 let cursor = match self.load_cursor(name, self.size) {
                     None => {
-                        let Some(ref fallback) = self.fallback else {
-                            return None;
-                        };
+                        let fallback = self.fallback.as_ref()?;
                         let data = fallback.0(name, self.size)?;
                         let images = xparser::parse_xcursor(&data)?;
                         let conn = Connection::from_backend(self.backend.upgrade()?);
@@ -217,14 +217,15 @@ impl CursorTheme {
     /// array with the contents of an `xcursor` file, or `None` if you don't provide a fallback for this cursor.
     ///
     /// For example, this defines a generic fallback cursor image and uses it for all missing cursors:
-    /// ```
+    /// ```ignore
     /// # use wayland_cursor::CursorTheme;
     /// # use wayland_client::{Connection, backend::InvalidId, protocol::wl_shm};
     /// # fn example(conn: &Connection, shm: wl_shm::WlShm, size: u32) -> Result<CursorTheme, InvalidId> {
-    /// let mut theme = CursorTheme::load_or(conn, shm, "default", size);
-    /// theme.set_callback(|name, size| {
-    ///     include_bytes!("./icons/default")
-    /// });
+    /// #   let mut theme = CursorTheme::load_or(conn, shm, "default", size)?;
+    /// #   theme.set_callback(|name, size| {
+    /// #       include_bytes!("./icons/default")
+    /// #   });
+    /// #   Ok(theme)
     /// # }
     /// ```
     pub fn set_callback<F>(&mut self, fallback: F)
--- a/wayland-cursor/src/lib.rs
+++ b/wayland-cursor/src/lib.rs
@@ -45,7 +45,9 @@
 //! # }
 //! ```
 
+use std::borrow::Cow;
 use std::env;
+use std::fmt::Debug;
 use std::fs::File;
 use std::io::{Error as IoError, Read, Result as IoResult, Seek, SeekFrom, Write};
 use std::ops::{Deref, Index};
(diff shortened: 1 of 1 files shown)


elinorbgr:
Thanks, that is interesting, however I'm not sure this design is the whole way we want to go.

As I understand, the goal is to allow an app to provide fallback cursor images by itself in case the system theme does not provide all the ones it needs.

The method you introduce directly creates a `Cursor` instance, storing its contents using the SHM pool of the theme, but without any other checks, nor inserting it into the internal state. As a result the user will have to keep the `Cursor` instance around themselves (as the `CursorTheme` does not store it) and check to use it whenever they need this particular cursor, to avoid reloading a new copy in memory every time.

I think a better approach would be to introduce a proper mechanism by which the user of `wayland-cursor` can specify a general fallback mechanism. A possible implementation of that would be to let the user provide a fallback `Fn(&str) -> Cow<'static, [u8]>` closure that would be invoked by `load_cursor` if the requested cursor was not found in the system theme. (The `Cow<'static, [u8]>` return type being used so that this closure could either return a reference to some cursor data bundled in the executable itself, or load the contents from disk in the app files.)

With such a fallback mechanism, later use of the `CursorTheme` would be transparent, and everything would "just work".

What do you think of this, would that cover your needs?

wayland-cursor/src/lib.rs
```diff
@@ -177,6 +194,63 @@ impl CursorTheme {
...
     }
 
+    /// try to load data from theme, else, it will try to load from fallback data
+    ///
+    /// This method returns [`None`] if this cursor is not provided either by the theme, or by one of its parents,
+    /// and the data by fallback function cannot parse it
+    /// fallback function will receive a enum which discribe the error, you can print some messages
+    /// or do some other things before return the finally fallback data
+    pub fn get_cursor_with_default_data<F>(&mut self, name: &str, fallback: F) -> Option<&Cursor>
+    where
```
elinorbgr, line 204: I was rather thinking of defining the fallback closure one and for all as a standalone method of `CursorTheme`, so that the rest of the API does not need to change at all.

Or is having an independent closure for each time you retrieve the cursor, with all that error-processing machinery something that you need for your uses?

Decodetalkers: Emm, what do you mean , make the function out of CursorTheme?

elinorbgr: My idea was to add to `CursorTheme` a method like this:

```rust
pub fn set_fallback(fb: impl FnMut(&str) -> Cow<'static, [u8]>) { /* ... */ }
```

that stores a fallback closure inside the `CursorTheme`, that would then be used internally by `get_cursor()` if the system theme does not provide the requested cursor.

Decodetalkers: then I think pass a fb is needless, and maybe here need to return a result

Decodetalkers: ha, I see

Shinyzenith:
Hi gentle ping: @elinorbgr Can we get a review on this? 😄

elinorbgr (commented):
Sorry for the long time, this kinda slipped outside of my radar...

LGTM overall, but:

- the `Fallback` function should take a size as argument in addition to the cursor name, like the `load_cursor`, making it possible to have multiple sizes for the fallback cursor if desired
- Could you expand the documentation of `set_fallback()` to make it clearer what that callback is supposed to do, and how it's meant to be used ?

Decodetalkers:
> Sorry for the long time, this kinda slipped outside of my radar...
> 
> LGTM overall, but:
> 
>     * the `Fallback` function should take a size as argument in addition to the cursor name, like the `load_cursor`, making it possible to have multiple sizes for the fallback cursor if desired
> 
>     * Could you expand the documentation of `set_fallback()` to make it clearer what that callback is supposed to do, and how it's meant to be used ?

Ok, does that be ok? I have push another comment

wayland-cursor/src/lib.rs
```diff
@@ -160,23 +180,57 @@ impl CursorTheme {
...
+                };
                 self.cursors.push(cursor);
                 self.cursors.iter().last()
             }
         }
     }
 
+    /// Set the callback to load the data, the return value of the fallback function should be the source of
+    /// xcursor
+    /// Same as calling the following:
```
elinorbgr, line 216:
```suggestion
    /// Set a callback to load the cursor data, in case the system theme is missing a cursor that you need.
    ///
    /// Your callback will be invoked with he name and size of the requested cursor and should return a byte
    /// array with the contents of an `xcursor` file, or `None` if you don't provide a fallback for this cursor.
    ///
    /// For example, this defines a generic fallback cursor image and uses it for all missing cursors:
```

Decodetalkers:
done @elinorbgr

elinorbgr (approved):

wayland-cursor/src/lib.rs
```diff
@@ -160,23 +182,59 @@ impl CursorTheme {
...
+    /// # use wayland_client::{Connection, backend::InvalidId, protocol::wl_shm};
+    /// # fn example(conn: &Connection, shm: wl_shm::WlShm, size: u32) -> Result<CursorTheme, InvalidId> {
+    /// #   let mut theme = CursorTheme::load_or(conn, shm, "default", size)?;
+    /// #   theme.set_callback(|name, size| {
+    /// #       include_bytes!("./icons/default")
+    /// #   });
+    /// #   Ok(theme)
+    /// # }
+    /// ```
+    pub fn set_callback<F>(&mut self, fallback: F)
```
MarijnS95, line 231: While patching up the docs, is it intentional that this is called `set_callback()` and not `set_fallback()` (where the _fallback_ is a _callback_ that provides a _fallback_ value when called)?

Decodetalkers: oh.. it is a mistake. can you make a pr for me? thanks

Decodetalkers: `set_fallback`, it was the author suggested. but then it become a mistake

MarijnS95: @Decodetalkers done in #723!

PR #682, opened 2023-12-17
client: Include version info and interface name in `BindError`

MarijnS95:
When `BindError` is returned (and bubbled up by callers) it is no longer clear what interface or version was requested (nor what is the highest this server might be able to provide), leaving users to wonder what the context for the error is.

Add this information to the object to elaborate in error messages and allow callers to read the values after the fact.

---

This supersedes #671 by also including the requested version and interface name (in addition to the available version), and a documtation fix.

Note that all of this might be considered redundant as the caller typically knows these values (they passed them as function arguments), and it would be up to them to re-wrap the error with this info as context?  Is there a policy for this?

### DIFF
--- a/wayland-client/src/globals.rs
+++ b/wayland-client/src/globals.rs
@@ -135,7 +135,7 @@ impl GlobalList {
     /// version of the returned protocol object is the lower of the maximum requested version and the advertised
     /// version.
     ///
-    /// If the lower bound of the `version` is less than the version advertised by the server, then
+    /// If the lower bound of the `version` is greater than the version advertised by the server, then
     /// [`BindError::UnsupportedVersion`] is returned.
     ///
     /// ## Multi-instance/Device globals.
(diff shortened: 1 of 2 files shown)


wayland-client/src/globals.rs
```diff
@@ -115,7 +115,7 @@ impl GlobalList {
     /// version of the returned protocol object is the lower of the maximum requested version and the advertised
     /// version.
     ///
-    /// If the lower bound of the `version` is less than the version advertised by the server, then
+    /// If the lower bound of the `version` is greater than the version advertised by the server, then
     /// [`BindError::UnsupportedVersion`] is returned.
```
MarijnS95, line 119: What I assume happened here is that `version` in the code is the bound version of what is found in the global registry (i.e. the version that is _available_), while "the lower bound of the `version`" is rebond to `version_start`.

elinorbgr (approved):
LGTM thanks, but that PR needs a rebase it appears.

MarijnS95:
Yup, I was using this on top of `wayland-backend-v0.3.2` to test as `[patch.crates-io]` without running into incompatible changes. Should be good now :)

MarijnS95:
Thanks for the rebase and merge @ids1024!

PR #683, opened 2023-12-17
Update per-crate licenses

alebastr:
Several new crates were published without a mandatory MIT copyright notice, and that fails the checks in our downstream (Fedora) packaging tools.

See also #323


kchibisov:
I'd suggest to just adding symlinks, the way it's done with any other project having crates in workspaces.

i509VCB:
I've seen this type of issue at least enough times to question whether there should be a clippy lint (workspace wide) to warn about this

alebastr:
> I'd suggest to just adding symlinks, the way it's done with any other project having crates in workspaces.

Fixed. I assumed there was a reason to copy the files initially. If there is none, symlinks indeed are better.

> I've seen this type of issue at least enough times to question whether there should be a clippy lint (workspace wide) to warn about this

That would require being able to detect what files contain copyright notices. IIRC, the cargo team was against using `license-files` alongside the `license` metadata attribute and heuristics makes a bad lint.
Besides, not every license requires distributing the full text along with the source. Just the most commonly used ones :)

NyxAlexandra:
IIRC there's a Cargo manifest option to specify the path to the license file. Would that be sufficient?

elinorbgr:
Thanks a lot!

elinorbgr:
IIRC at some point `cargo publish` had trouble with symlinks, I assume this is no longer an issue?

kchibisov:
>IIRC at some point cargo publish had trouble with symlinks, I assume this is no longer an issue?

It's not an issue for a long time, I think, since that's how we're doing so in alacritty for a long time. But I do recall that it probably was at some point.

Issue #665, opened 2023-10-02
SEGFAULT with `client_system` and `WAYLAND_DEBUG=1` using object in argument after calling `.destroy()`


ids1024:
Reproduction:
* Modify `wayland-client/examples/simple_window.rs` to add `buffer.destroy()` before `surface.attach(...)`
* Run with `WAYLAND_DEBUG=1 cargo run --example simple_window --features wayland-backend/client_system`
Result:
```
[...]
[4191435.215]  -> wl_buffer@7.destroy()
[4191435.228]  -> wl_surface@4.attach(Segmentation fault (core dumped)
```

Presumably this issue is not specific `surface.attach` this call, but it happens to be where I've seen it.

PR #685, opened 2023-12-19
Fix test of `alive` when sending object argument

ids1024:
Fixes https://github.com/Smithay/wayland-rs/issues/665.


elinorbgr:
LGTM thanks, could you just add a changelog entry for this please?

wayland-backend/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 ### Additions
 - client: Implement `Eq` for `Backend`
 
+#### Bugfixes
+- backend: Fix error/segfault if object argument is no longer alive
```
elinorbgr, line 9:
```suggestion
- sys/client: Fix error/segfault if object argument is no longer alive
```

BTW, that makes me realize, it's likely there is a similar bug server-side, that'd be better to fix both in the same PR is that is the case.

elinorbgr (approved):

PR #693, opened 2024-01-25
Generate aliases with "z" stripped for stable dmabuf protocol

PolyMeilex:
As you may know, `zwp_linux_dmabuf_v1` became stable without striping of the `z` prefix.
This is just a random idea to do the striping ourselves, by aliasing all `v1` types to `zv1` ones. 

Eg.
- wp_linux_buffer_params_v1
  - WpLinuxBufferParamsV1 -> ZwpLinuxBufferParamsV1
  - Event -> Event
  - Request -> Request
  - ...

### DIFF
--- a/wayland-protocols/src/wp.rs
+++ b/wayland-protocols/src/wp.rs
@@ -179,6 +179,24 @@ pub mod linux_dmabuf {
             []
         );
     }
+
+    /// Stable version 1 (Aliased to zv1, as zv1 became stable without name change)
+    pub mod v1 {
+        #[cfg(feature = "client")]
+        pub mod client {
+            //! Client-side API of this protocol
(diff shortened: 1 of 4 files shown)


elinorbgr:
I fear this would create confusion more than anything else... We'd end up having names that no longer match the rest of the Wayland ecosystem, nor the contents of the debug logs.

ids1024:
It's kind of a strange situation. On the one hand, `zwp_linux_dmabuf_v1` *is* the name of the protocol. On the other hand, it *should* be named `wp_linux_dmabuf_v1` but just isn't for awkward technical reasons.

If we did this, it might be good to mark the old name as  `#[deprecated]`. Rather than just having duplicating in the API.

But overall, it may be best just to leave it as is.

PR #697, opened 2024-02-07
Don't try to connect to empty WAYLAND_SOCKET

kchibisov:
--

I basically moved code around so it looks a bit nicer.

### DIFF
--- a/wayland-client/src/conn.rs
+++ b/wayland-client/src/conn.rs
@@ -10,6 +10,8 @@ use std::{
     },
 };
 
+use rustix::io::{self as rustix_io, FdFlags};
+
 use wayland_backend::{
     client::{Backend, InvalidId, ObjectData, ObjectId, ReadEventsGuard, WaylandError},
     protocol::{ObjectInfo, ProtocolError},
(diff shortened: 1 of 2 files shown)


kchibisov:
The thing about it is that `libwayland` is not really doing that sort of stuff and empty `WAYLAND_SOCKET` will still prevent it from running.

wayland-client/src/conn.rs
```diff
@@ -44,48 +46,61 @@ impl Connection {
...
+                return Err(ConnectError::InvalidFd);
+            }
+        };
+
+        let backend = Backend::connect(stream).map_err(|_| ConnectError::NoWaylandLib)?;
+        Ok(Self { backend })
+    }
+
+    /// Try to connect with WAYLAND_DISPLAY env.
+    pub fn connect_to_wayland_display() -> Result<Self, ConnectError> {
```
i509VCB, line 83: I question whether we want this to be public, as if the server is giving you an explicit socket, you should probably use it. Or is this a case where winit may want to fallback if the socket is garbage?

Otherwise the name is a possibly confusing, someone may think you use this function to connect to a display, but not realize their application should just use the env.

i509VCB:
> The thing about it is that `libwayland` is not really doing that sort of stuff and empty `WAYLAND_SOCKET` will still prevent it from running.
> 
> So maybe all of that is actually intended?

Probably worth an issue upstream with libwayland

wayland-client/src/conn.rs
kchibisov: Oh, it should be private.

kchibisov:
https://gitlab.freedesktop.org/wayland/wayland/-/issues/438

elinorbgr:
I guess we will keep following upstream behaviour on this one, for consistency.

Issue #698, opened 2024-02-11
`it is undefined behavior to use this value` error on latest nightly compiler


doonv:
Building on the latest nightly compiler (rustc 1.78.0-nightly (6cc484351 2024-02-10)) causes the following error:

```
   Compiling wayland-client v0.31.2
error[E0080]: it is undefined behavior to use this value
   --> ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wayland-client-0.31.2/src/lib.rs:218:9
    |
218 |         wayland_scanner::generate_interfaces!("wayland.xml");
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ constructing invalid value at .<deref>[1].child_interface.<enum-variant(Some)>.0.<deref>.c_ptr.<enum-variant(Some)>.0: encountered reference to mutable memory in `const`
    |
    = note: The rules on what exactly is undefined behavior aren't clear, so this check might be overzealous. Please open an issue on the rustc repository if you believe it should not be considered undefined behavior.
    = note: the raw bytes of the constant (size: 8, align: 8) {
                ╾───alloc2328<imm>────╼                         │ ╾──────╼
            }
    = note: this error originates in the macro `wayland_scanner::generate_interfaces` (in Nightly builds, run with -Z macro-backtrace for more info)

note: erroneous constant encountered
   --> ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wayland-client-0.31.2/src/lib.rs:218:9
    |
218 |         wayland_scanner::generate_interfaces!("wayland.xml");
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this note originates in the macro `wayland_scanner::generate_interfaces` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0080]: it is undefined behavior to use this value
   --> ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wayland-client-0.31.2/src/lib.rs:218:9
    |
218 |         wayland_scanner::generate_interfaces!("wayland.xml");
[...]

PR #699, opened 2024-02-11
Delegated dispatchers

PolyMeilex:
This is an attempt to remove all delegate macros from Smithay, as well as remove obnoxious type constraints on `D`.

Here is an example diff of `ZxdgOutputManagerV1` implementation in Smithay with new delegated dispatches.

```diff
impl<D> GlobalDispatch<ZxdgOutputManagerV1, (), D> for OutputManagerState
where
-    D: GlobalDispatch<ZxdgOutputManagerV1, ()>,
-    D: Dispatch<ZxdgOutputManagerV1, ()>,
-    D: Dispatch<ZxdgOutputV1, XdgOutputUserData>,
-    D: 'static,
{
}
```

All of the constraints on `D` can be removed, as `OutputManagerState` is the one implementing them.

This also completely removes the need to write and invoke `delegate_*!` macros.

I managed to implement delegated `GlobalData` without any breaking changes just by adding new default generic, but for `ResourceData` it's more cumbersome, as I had to add new `DelegatedResourceData` user data type. While this does not introduce any braking changes, the split is weird. As we now have `.data::<U>()` and `.delegated_data::<U, DelegatedTo>()` variants. Ideas on how to merge those types together without polluting the whole API are welcome.

(Smithay PR that demonstrates this in action https://github.com/Smithay/smithay/pull/1327)


elinorbgr:
This is a nice take on this problem! Thanks a lot, the general principle looks good to me :+1:

Given this alters the generated code in a non-backward-compatible way, this will be a breaking change of wayland-scanner, and I think we should probably treat it as a breaking change of wayland-server as well as remove the cumbersome macros in the process.

Apart from waiting on the resolution of https://github.com/Smithay/wayland-rs/issues/698 for our CI to be functional again, I think before this is in a mergeable state it needs:
- Proper documentation of the newly introduced API, and update of the relevant module-level docs
- Equivalent changes for wayland-client
- Proper changelog entries

ids1024:
For an implementation in `wayland-client`, will this run into issues with requests that take a `new_id`?

https://docs.rs/wayland-client/latest/wayland_client/protocol/wl_seat/struct.WlSeat.html#method.get_keyboard for instance is defined as:

```rust
pub fn get_keyboard<U: Send + Sync + 'static, D: Dispatch<WlKeyboard, U> + 'static>(
    &self,
    qh: &QueueHandle<D>,
    udata: U
) -> WlKeyboard
```

Without the delegated implementation the macro generated, `Dispatch<WlKeyboard, U>`  wouldn't be implemented.

Events with a `new_id` meanwhile are rare, and it seems `wayland-server` handles these differently (Smithay has to use `backend.create_object` directly.) It seems a bit odd that wayland-server and wayland-client do this differently.

PolyMeilex:
Ok, client side is done.
I went ahead and broke the API of `QueueHandle::make_data<_,_, DelegateTo>`, by adding the `DelegateTo`, as I suppose it's rarely used anyway, and this allows us to avoid delegated/non-delegated split of the API.

Code generated by the scanner will just output:
```rs
fn get_registry(qh) -> WlRegistry {
    let data = qh.make_data::<_, _, D>();
    ...
}
```
So nothing will change for the general case of global `D` type.

We as SCTK on the other hand can call the constructor manually with our `DelegateTo` type instead:
```rs
let data = qh.make_data::<WlRegistry, _, DelegateTo>(());
let wl_registry = display.send_constructor(wl_display::Request::GetRegistry {}, data).unwrap();
```

I belive that the code needed to call constructors manually is simple enough to no warrant an additional `get_registry_delegated::<DelegateTo>()` generation, or forcing the `DelegateTo` generic on regular `get_registry`.

---

Also it turns out that thanks to [ObjectData::data_as_any](https://docs.rs/wayland-backend/latest/wayland_backend/sys/client/trait.ObjectData.html#method.data_as_any) on client side the generic on data getter is totally not needed, I will experiment with similar approach on server side as well.

ids1024:
> I belive that the code needed to call constructors manually is simple enough to no warrant an additional get_registry_delegated::<DelegateTo>() generation, or forcing the DelegateTo generic on regular get_registry.

Yeah, we wouldn't really want to end up with a duplicated `_delegated` method for every method that creates an object. Not to require awkward explicit type generics. (And we can't have a generic with a default on methods: https://github.com/rust-lang/rust/issues/36887).

The only other possibility I can think of is if `Dispatch` was implemented on `UserData`, not `State`. Then the dispatcher would be implied by the type of the `udata`, at the cost of not being able to use standard types like `()` as a udata. This would also be a bigger breaking change.

PolyMeilex:
> Also it turns out that thanks to [ObjectData::data_as_any](https://docs.rs/wayland-backend/latest/wayland_backend/sys/client/trait.ObjectData.html#method.data_as_any) on client side the generic on Proxy::data getter is totally not needed, I will experiment with similar approach on server side as well.

I gave up, the `D` generic on server side `ObjectData<D>` trait makes it harder to do, especially with`trait_upcasting` being nightly only, so I'll leave it as is.

Btw, Docs should be up-to-date now as well. So I guess this could be considered ready.

PolyMeilex:
Gentle ping @elinorbgr, no rush tho.

wayland-client/src/lib.rs
```diff
@@ -42,10 +42,13 @@
 //! managing the newly created object.
 //!
 //! However, implementing all those traits on your own is a lot of (often uninteresting) work. To make this
-//! easier a composition mechanism is provided using the [`delegate_dispatch!`] macro. This way, another
-//! library (such as Smithay's Client Toolkit) can provide generic [`Dispatch`] implementations that you
+//! easier a composition mechanism is provided. Using the [`QueueHandle<State>::make_data()`] method
+//! one is able to create object handlers manually, making it possible to specify `DispatchTo` type diferent
+//! then your `State`.
```
elinorbgr, line 47:
```suggestion
//! one is able to create object handlers manually, making it possible to specify `DispatchTo` type different
//! from your `State`.
```

wayland-client/src/lib.rs
```diff
@@ -42,10 +42,13 @@
...
+//! Library (such as Smithay's Client Toolkit) can use that mechanism to provide generic [`Dispatch`] implementations that you
```
elinorbgr, line 48:
```suggestion
//!
//! Libraries (such as Smithay's Client Toolkit) can use that mechanism to provide generic [`Dispatch`] implementations that you
```

wayland-client/src/lib.rs
```diff
@@ -42,10 +42,13 @@
...
 //! can reuse in your own app by delegating those objects to that provided implementation. See the
-//! documentation of those traits and macro for details.
+//! documentation of [`QueueHandle<State>::make_data()`] for details.
+//! There is also `delegated.rs` example in `wayland-rs/wayland-client`.
```
elinorbgr, line 51: Maybe lets make this a  direct link to where the example file will be once this PR is merged?

wayland-client/src/event_queue.rs
```diff
@@ -604,14 +604,16 @@ impl<State: 'static> QueueHandle<State> {
     /// This creates an implementation of [`ObjectData`] fitting for direct use with `wayland-backend` APIs
     /// that forwards all events to the event queue associated with this token, integrating the object into
     /// the [`Dispatch`]-based logic of `wayland-client`.
-    pub fn make_data<I: Proxy + 'static, U: Send + Sync + 'static>(
+    ///
+    /// Events will be dispatched via [`Dispatch`] to a `DelegateTo` generic type.
```
elinorbgr, line 608: I have a feeling that this is a bit too succinct, but I'm not quite sure how to expand it.

wayland-server/src/display.rs
```diff
@@ -134,7 +134,34 @@ impl DisplayHandle {
...
+        )
+    }
+
+    /// Create a new protocol global
+    ///
+    /// This global will be advertized to clients through the `wl_registry` according to the rules
+    /// defined by your [`GlobalDispatch`] implementation for the given interface. Whenever a client
+    /// binds this global, the associated [`GlobalDispatch::bind()`] method will be invoked on your
+    /// `State`.
+    pub fn create_delegated_global<State, I, U, DelegateTo>(
```
elinorbgr, line 147: This documentation was copy/pasted without being updated

elinorbgr (changes requested):
LGTM overall, a few documentation nitpicks and two last points:

- Please add changelog entries for these changes to the relevant crates (wayland-client, wayland-server and wayland-scanner)
- The two new examples could really use a few comments that explain what they do, what the flow of messages will be, etc...

wayland-client/src/lib.rs
PolyMeilex: Done in ee9140d38bbf6922653196a3fbe1200b175808cc

wayland-client/src/lib.rs
PolyMeilex: Done in ee9140d38bbf6922653196a3fbe1200b175808cc

wayland-client/src/lib.rs
PolyMeilex: Done https://github.com/Smithay/wayland-rs/pull/699/commits/ee9140d38bbf6922653196a3fbe1200b175808cc#diff-82d250d3dc7f4360cf4c5a8a0f88f1c0fb30bce47219bcbbdcd9ef40a74d732fR51

wayland-client/src/event_queue.rs
PolyMeilex: Attempted to clarify that by adding a pseudocode example https://github.com/Smithay/wayland-rs/pull/699/commits/ee9140d38bbf6922653196a3fbe1200b175808cc#diff-ce00ed5e221ce826207c7b1cff410364eeb6e30c9322e7a875129b8a50871198R609-R627

wayland-server/src/display.rs
PolyMeilex: My bad, addressed in https://github.com/Smithay/wayland-rs/pull/699/commits/ee9140d38bbf6922653196a3fbe1200b175808cc#diff-5e5b2f5127938323ccf49f8f4096fe02d2f630fe4338a4b345a79d42ff941a11R143-R147

wayland-client/src/event_queue.rs
```diff
@@ -606,6 +606,25 @@ impl<State: 'static> QueueHandle<State> {
...
+    ///
+    /// struct SeatDispatcher;
+    ///
+    /// impl<D> Dispatch<WlSeat, (), D> for SeatDispatcher {
+    ///     // if `DelegateTo` of an object is set to `SeatDispatcher` events will be dispatched here
+    ///     // to this impl
+    /// }
+    ///
+    /// let obj1 = qh.make_data<WlOutput, (), OutputDispatcher>;
+    /// let obj2 = qh.make_data<WlSeat, (), SeatDispatcher>;
```
elinorbgr, line 626: I don't think this is a valid rust syntax

elinorbgr (commented):
Aren't we also removing the delegate_dispatch macros in the process? Or do you think there is a good reason to keep them alongside this new API?

PolyMeilex:
> Or do you think there is a good reason to keep them alongside this new API?

Nope, removed.
Those are easy to copy-paste if someone really needs them for some reason.

ids1024:
In wayland-client, `Globals::bind` still has a bound on `State: Dispatch<I, U> + 'static`. Presumably it needs to be changed/duplicated to support delegation?

ids1024:
Seems to work with something like:

```rust
    pub fn bind_delegated<I, State, U, DelegateTo>(
        &self,
        qh: &QueueHandle<State>,
        version: RangeInclusive<u32>,
        udata: U,
    ) -> Result<I, BindError>
    where
        I: Proxy + 'static,
        State: 'static,
        U: Send + Sync + 'static,
        DelegateTo: Dispatch<I, U, State> + 'static,
    {
        ...

        Ok(self
            .registry
            .send_constructor(
                wl_registry::Request::Bind { name, id: (I::interface(), version) },
                qh.make_data::<I, U, DelegateTo>(udata),
            )
            .unwrap())
```

Then `bind` can be defined as:

```rust
    pub fn bind<I, State, U>(
        &self,
        qh: &QueueHandle<State>,
        version: RangeInclusive<u32>,
        udata: U,
    ) -> Result<I, BindError>
    where
        I: Proxy + 'static,
        State: Dispatch<I, U> + 'static,
        U: Send + Sync + 'static,
    {
        self.bind_delegated::<I, State, U, State>(qh, version, udata)
    }
```

ids1024:
`registry_queue_init` likewise requires `State: Dispatch<wl_registry::WlRegistry, GlobalListContents>`. So without sctk's `relegate_registry`, this now requires the application using sctk to explicitly implement `Dispatch<wl_registry::WlRegistry, GlobalListContents>`.

ids1024:
https://github.com/Smithay/wayland-rs/pull/902 has been merged as part of the next major release (https://github.com/Smithay/wayland-rs/issues/900), so this has been superseded with hopefully a better variation on this.

PR #700, opened 2024-02-12
Use a wrapper type implementing `Sync`, instead of `static mut`

ids1024:
`static mut` is generally best avoided. It seems to only be necessary currently to allow the use of non-`Sync` types, since pointers are not sync.

This also removes the `NULLPTR` constant in favor of using `std::ptr::null`, uses `.as_ptr()` instead of casting, and uses the arrray repeating syntax to define `types_null`. None of those should impact behavior or the public API.

`*_requests` and `*_events` static are no longer public now. This is theoretically a breaking change, but shouldn't really impact anything.

### DIFF
--- a/wayland-scanner/src/c_interfaces.rs
+++ b/wayland-scanner/src/c_interfaces.rs
@@ -1,5 +1,4 @@
 use std::cmp;
-use std::iter::repeat;
 
 use proc_macro2::{Literal, TokenStream};
 use quote::{format_ident, quote};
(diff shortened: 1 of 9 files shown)


ids1024:
Re-ran CI so it would use the new nightly, and everything passes.

wayland-scanner/src/c_interfaces.rs
```diff
@@ -27,14 +26,13 @@ pub(crate) fn generate_interfaces_prefix(protocol: &Protocol) -> TokenStream {
...
-        .take(longest_nulls);
-
     quote! {
-        const NULLPTR: *const std::os::raw::c_void = 0 as *const std::os::raw::c_void;
-        static mut types_null: [*const wayland_backend::protocol::wl_interface; #types_null_len] = [
-            #(#nulls,)*
-        ];
+        use std::ptr::null;
+        struct SyncWrapper<T>(T);
+        unsafe impl<T> Sync for SyncWrapper<T> {}
```
elinorbgr, line 32: Shouldn't this have a `#[repr(C)]` or something like that for good measure?

ids1024: The code here uses `.0.as_ptr()` to get a pointer to the array wrapped in the `SyncWrapper`. So the `repr` of the `SyncWrapper` type shouldn't matter.

elinorbgr: Oh right, that's good then. :+1:

elinorbgr:
Could you just add a changelog entry for the scanner for this? AFAICT this is not a breaking change, but it's still noteworthy imo.

ids1024:
Mentioned in changelog, and fixed more CI warnings.

PR #707, opened 2024-03-17
wayland-backend: raw-window-handle 0.6

i509VCB:
Unfortuntely we used the `raw-window-handle` feature name for 0.5, which would be a breaking change. To handle this for the future, replicate what winit does and make a new feature for 0.6. The 0.5 version feature is deprecated and will be removed in a later version.

### DIFF
--- a/wayland-backend/Cargo.toml
+++ b/wayland-backend/Cargo.toml
@@ -19,6 +19,7 @@ log = { version = "0.4", optional = true }
 scoped-tls = "1.0"
 downcast-rs = "1.2"
 raw-window-handle = { version = "0.5.0", optional = true }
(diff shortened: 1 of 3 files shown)


elinorbgr:
The CI failure looks unrelated, could you just add a changelog entry for this as well please?

elinorbgr:
Actually, the fail of the of "format" job is legit, looks like a trailing space slipped in.

ids1024:
This still needs a changelog entry and formatting fix. But otherwise seems good.

ids1024:
Rebased and merged as part of https://github.com/Smithay/wayland-rs/pull/732.

PR #711, opened 2024-04-04
wayland-client: add example for bypassing Dispatch

Supreeeme:
Also mention ObjectData in the docs for bypassing Dispatch.


elinorbgr (approved):
LGTM, Thanks a lot!

PR #715, opened 2024-04-19
Headerless xml parsing fix

lmaxyz:
Found out some protocols are not available because of parser.
That patch fixes parsing for headerless xml files.

Also I added couple of tests.

### DIFF
--- a/wayland-protocols-plasma/CHANGELOG.md
+++ b/wayland-protocols-plasma/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Make available protocols that based on headerless xml files.
+
 ## 0.2.0 -- 2023-09-02
 
 ### Breaking changes
--- a/wayland-scanner/CHANGELOG.md
+++ b/wayland-scanner/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Use wrapper type implementing `Sync` instead of `static mut`s.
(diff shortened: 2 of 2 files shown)


wayland-scanner/tests/scanner_assets/test-headerless-protocol.xml
```diff
@@ -0,0 +1,252 @@
+<protocol name="test-protocol">
```
elinorbgr, line 1: Given this test file is only used to ensure the scanner doesn't choke on it, I don't think it needs to be as big as the main one, it can be much more minimalistic I'd say.

lmaxyz: It's done, thanks.

elinorbgr:
LGTM, could you just add an entry to the changelogs of wayland-scanner and wayland-protocols-plasma to reflect those changes ?

Also, it looks like rustfmt is unhappy in the CI

lmaxyz:
> LGTM, could you just add an entry to the changelogs of wayland-scanner and wayland-protocols-plasma to reflect those changes ?
> 
> Also, it looks like rustfmt is unhappy in the CI

Fixed fmt issues and added changes to changelogs.

lmaxyz:
@elinorbgr can you check it pls? Is it ok now?

elinorbgr (approved):
LGTM yes, thanks!

PR #716, opened 2024-04-25
Fix redundant/broken intradoc links and minor cleanup

MarijnS95:
After spotting a missing closing parenthesis in ``[`EventQueue::blocking_dispatch()`](EventQueue::blocking_dispatch``, leading to the link not rendering correctly, I set out on a course to delete these unnecessary link targets as ``[`EventQueue::blocking_dispatch()`]`` on its own already translates to the expected doc link, without being verbose. At the same time this PR includes a smattering of improvements, such as fixing missing links and updating mentions of functions that have been renamed or moved.

Finally, this PR sets up an extra CI step to make sure documentation doesn't have any build errors going forward.

I don't think this PR should be merged as-is, before we set up a little style guide for docs and finish the remainder. There's a bit of a mix currently but nothing we can't fix, so I'd like to ask the maintainers for their preference/agreement:

- Convert all `[]()` links to `[][]`, as the former converts into a nonsense URL sometimes when the target is not a valid item and there's no matching `/// [bar]: ...` label target in scope, whereas the latter fails the build appropriately (as described by me in https://github.com/rust-windowing/winit/pull/3178#discussion_r1371704466);
- Links to functions should use `()` in the name to avoid ambiguity;
- Link text that points to a different destination (i.e. ``[`foo()`][Self::foo()]``) should use parenthesis in the target as well, again to avoid ambiguity;
- Links to functions on the current "scope" i.e. on `Self`, should be named as:
  - ``[`FullStructName::func()`]``;
  - ``[`func()`][FullStructName::func()]``;
  - ``[`func()`][Self::func()]``;
- Links with `crate` in them (like ``[`crate::event_created_child!()`]``) should be avoided and instead use an explicit link target.

Thanks for considering this! The documentation is already great and I hope this makes it more consistent (and less likely to reference (re)moved items) going forward!


elinorbgr:
Thanks a lot for this work, your suggestions seem quite good. :+1:

I honestly don't have any strong feelings about the precise documentation style we use, but I agree that having a consistent one would absolutely be a good thing. Do you have any thoughts on that @ids1024 ?

MarijnS95:
Thanks @elinorbgr! Since proposing that list I've made a few picks already, bit the bullet, and pushed the changes to consistenize the docs. Only removing `crate::` on macros remains, and do point it out to me if I forgot a link/comment or two.

There were a few non-doc `foo()` references already, there are probably more to find that should be turned into proper links before marking this as finished.

ids1024:
> Convert all []() links to [][], as the former converts into a nonsense URL sometimes when the target is not a valid item and there's no matching /// [bar]: ... label target in scope, whereas the latter fails the build appropriate

I didn't know `[][]` links like that were a thing, but that sounds like a good reason to use them. Broken links like that can be a common problem in Rust documentation.

ids1024 (approved):
The changes here all look good to me. I also like having CI check for `cargo doc` warnings like this.

.github/workflows/ci.yml
```diff
@@ -416,6 +416,8 @@ jobs:
       - name: Build Documentation
         uses: actions-rs/cargo@v1
         with:
+          env:
+            RUSTDOCFLAGS: -Dwarnings
```
elinorbgr, line 420: It looks like this change broke the CI script, it now fails to parse with the following error:

```
The workflow is not valid. .github/workflows/ci.yml (Line: 420, Col: 13): A mapping was not expected
```

MarijnS95: It shouldn't have been in the `with:` binding, and I hadn't seen that because the CI workflow requires approval to run (which is silly).

MarijnS95:
Since I have rebased, I should probably check if there are any new docs merged since that time that need touch-ups. And maybe set the "style guideline" in stone somewhere?

MarijnS95:
Looks like `nightly` in the CI immediately picked up on https://blog.rust-lang.org/2024/05/06/check-cfg.html, for which I've filed #722 to address this separately.

ids1024:
Presumably CI will pass if this is rebased.

MarijnS95:
@ids1024 it will, though I'd like https://github.com/Smithay/wayland-rs/pull/723 to be merged first to get rid of some of the unrelated changes in this PR.

wayland-client/src/event_queue.rs
```diff
@@ -30,14 +30,14 @@ use crate::{conn::SyncData, Connection, DispatchError, Proxy};
 ///
 /// In the rare case of an interface with *events* creating new objects (in the core protocol, the only
 /// instance of this is the `wl_data_device.data_offer` event), you'll need to implement the
-/// [`Dispatch::event_created_child()`] method. See the [`event_created_child!`](macro.event_created_child.html) macro
+/// [`Dispatch::event_created_child()`] method. See the [`crate::event_created_child!()`] macro
```
MarijnS95, line 33: Not sure if we want to get rid of `crate::` here and below?

ids1024: Yeah, displaying `crate::` in the documentation doesn't really make much sense. So using `[][]` to avoid display `crate::` seems like a good idea.

MarijnS95:
Looks like a local Rust 1.78 run shows quite a few more issues now :tada:

PR #720, opened 2024-05-05
wayland-server: alive check for Weak

cmeissl:
In `smithay` we quite extensively use `Weak` as the key for caching some other resource. Cleaning up these resources
often requires checking if the key is still alive. Without an explicit function to check `Weak` for alive we have to resort to `Weak::upgrade().is_ok()`. An explicit function can save us from actually upgrading the resource itself, saving us quite some cpu time.

### DIFF
--- a/wayland-backend/src/server_api.rs
+++ b/wayland-backend/src/server_api.rs
@@ -249,6 +249,7 @@ impl WeakHandle {
     /// Try to upgrade this weak handle to a [`Handle`]
     ///
     /// Returns `None` if the associated backend was already dropped.
+    #[inline]
     pub fn upgrade(&self) -> Option<Handle> {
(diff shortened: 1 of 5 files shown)


elinorbgr:
Could you split out the rustfmt changes into another PR? Looks like the `check-minimal` CI is unhappy about something so I'd prefer we split this to figure out what's going on.

Other than that, LGTM :+1:

ids1024:
https://github.com/Smithay/wayland-rs/pull/722 should fix the CI failure if this is rebased.

cmeissl:
sorry for late reply, ran out of time trying to fix the CI build. dropped the commit with the attempt to fix the CI as it is already fixed in master, thx :)

ids1024 (approved):

PR #723, opened 2024-05-07
cursor: Rename `set_callback()` to `set_fallback()` to match intent

MarijnS95:
CC @Decodetalkers based on discussions in https://github.com/Smithay/wayland-rs/pull/680#discussion_r1592467187.

The intent of this function is to set a _fallback_ that `get_cursor()` uses when no cursor can be loaded.  This fallback _is_ a _callback_ (or _closure_ / `Fn` in Rust terms) that calculates an optional fallback value based on the given name and size of the requested cursor.

Also make sure this "fallback" setter is properly linked from `get_cursor()`, making it easier to find when users attempt to understand the docs.


elinorbgr:
LGTM, thanks

PR #725, opened 2024-05-10
protocols: Mention breaking API change

ids1024:
Due to https://gitlab.freedesktop.org/wayland/wayland-protocols/-/commit/1c57b24ff867eabacb6ecc1e74b1e4d1ccafcf4b, the generated code here is now an enum instead of a `u32`.

It seems reasonable to make this a breaking change for `wayland-protocols-*`, but not release a semver bump of `wayland-client`/`wayland-server`.

### DIFF
--- a/wayland-protocols/CHANGELOG.md
+++ b/wayland-protocols/CHANGELOG.md
@@ -1,6 +1,12 @@
 # CHANGELOG: wayland-protocols
 
 ## Unreleased
+
+### Breaking changes
+- `set_constraint_adjustment`/`SetConstraintAdjustment` now takes a `ConstraintAdjustment` instead of a u32.
+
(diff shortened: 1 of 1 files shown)


wayland-protocols/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
     - `xdg-dialog-v1`
     - `linux-drm-syncobj-v1`
 
+### Breaking changes
+- `set_constraint_adjustment`/`SetConstraintAdjustment` now takes a `ConstraintAdjustment` instead of a u32.
```
elinorbgr, line 13: LGTM, but I try to keep the "breaking changes" section at the top of the changelog for each releases in general

PR #726, opened 2024-05-11
`wl_fixes` protocol

ids1024:
In addition to copying the protocol changes from https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/388, this required two things:

* Modifies `wayland-scanner` to generate a `WlRegistry` type server-side. Instead of trying to make it produce a valid `parse_request` to handle the special case of the `bind` request, it just uses `unimplemented!`.
* Adds a `destroy_object` method to destroy the object
  - Not sure if this is entirely correct as implemented.

See https://github.com/Smithay/smithay/pull/1428 for server-side use of this.


ids1024:
I guess the server should also remove the registry from `known_registries`  (looks like currently elements are only removed their when the client is gone.) Not sure if there's anything else that might need to be removed anywhere on registry destruction.

dcz-self:
I added some tests: https://github.com/Smithay/wayland-rs/pull/826

PR #727, opened 2024-05-15
backend: Add `destroy_object` method

ids1024:
This is needed for implementing the `wl_fixes` protocol. It should also make it possible to handle the requirement in `zwlr_output_configuration_v1::destroy` to destroy the `wlr_output_configuration_head`.

Split off from https://github.com/Smithay/wayland-rs/pull/726. See https://github.com/Smithay/smithay/pull/1428 and https://github.com/rust-windowing/softbuffer/pull/218 for server-side and client-side uses of this.

### DIFF
--- a/wayland-backend/src/rs/server_impl/client.rs
+++ b/wayland-backend/src/rs/server_impl/client.rs
@@ -99,6 +99,17 @@ impl<D> Client<D> {
         InnerObjectId { id, serial, client_id: self.id.clone(), interface }
     }
 
+    pub(crate) fn destroy_object(
+        &mut self,
+        id: InnerObjectId,
+        pending_destructors: &mut Vec<super::handle::PendingDestructor<D>>,
+    ) -> Result<(), InvalidId> {
+        let object = self.get_object(id.clone())?;
+        pending_destructors.push((object.data.user_data.clone(), self.id.clone(), id.clone()));
+        self.send_delete_id(id.clone());
+        Ok(())
+    }
+
     pub(crate) fn object_info(&self, id: InnerObjectId) -> Result<ObjectInfo, InvalidId> {
         let object = self.get_object(id.clone())?;
         Ok(ObjectInfo { id: id.id, interface: object.interface, version: object.version })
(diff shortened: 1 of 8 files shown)


wayland-backend/src/rs/server_impl/client.rs
```diff
@@ -199,7 +210,6 @@ impl<D> Client<D> {
 
         // Handle destruction if relevant
         if message_desc.is_destructor {
-            self.map.remove(object_id.id.id);
```
elinorbgr, line 202: why is this line removed?

wayland-backend/src/sys/client_impl/mod.rs
```diff
@@ -516,6 +516,37 @@ impl InnerBackend {
...
+                ffi_dispatch!(
+                    wayland_client_handle(),
+                    wl_proxy_set_user_data,
+                    id.id.ptr,
+                    std::ptr::null_mut()
+                );
+            }
+            alive.store(false, Ordering::Release);
+            udata.data.destroyed(id.clone());
+        }
```
elinorbgr, line 538: This should check if `id.id.alive` is not already `False` (like if the user tries to destroy the same object twice), and return `Err(InvalidId)` if it is, to match the behavior of the rs backend.

wayland-backend/src/sys/server_impl/mod.rs
```diff
@@ -565,6 +565,20 @@ impl InnerHandle {
...
+        let state = (&mut *state as &mut dyn ErasedState)
+            .downcast_mut::<State<D>>()
+            .expect("Wrong type parameter passed to Handle::destroy_object().");
+
+        PENDING_DESTRUCTORS.set(&(&mut state.pending_destructors as *mut _ as *mut _), || unsafe {
+            ffi_dispatch!(wayland_server_handle(), wl_resource_destroy, id.id.ptr);
+        });
+
+        Ok(())
+    }
```
elinorbgr, line 580: Same for this method, it should check if the provided `ObjectId` is still alive before destroying it.

elinorbgr (changes requested):
On the high-level principle LGTM, but see my comments for the implementation.

Also, I think we should add some tests to the wayland-backend test module to cover these API. I'm thinking at least the two cases : trying to destroy the same object twice, and trying to send a message through an object after it has already been manually destroyed.

wayland-backend/src/rs/server_impl/client.rs
ids1024: Unless I'm missing something, the line is redundant since `send_delete_id` also calls `self.map.remove`.

elinorbgr: Ah yes right, good point.

PR #712, opened 2024-04-17
In debug logging, print 'fixed' values as floats

colinmarc:
I got a bit confused by the following:

```
2024-04-17T19:51:21.930954Z DEBUG session{session_id=365288}: mmserver::compositor: wheel delta: 0.0 -48.0
[ 768472.301][rs] -> wl_pointer@15.axis_source(2)
[ 768472.333][rs] -> wl_pointer@15.axis(13025, 0, -12288)
```

PR #732, opened 2024-05-30
Updates for next release

ids1024:
Includes https://github.com/Smithay/wayland-rs/pull/707 (rebased and with formatting fix) and https://github.com/Smithay/wayland-rs/pull/712, as well as changelog entries. (And a dev dependency bump.)

Want to get these merged before a release. Creating a PR mainly to verify CI passes.

### DIFF
--- a/wayland-backend/CHANGELOG.md
+++ b/wayland-backend/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+### Additions
+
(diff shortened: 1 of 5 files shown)


ids1024:
Huh, seems updating `env_logger` runs into some weird bug in older versions of cargo? https://github.com/rust-lang/mdBook/issues/2318

I guess that can be dropped.

PR #735, opened 2024-06-08
wayland-backend: Add getter for global name

PolyMeilex:
This will be needed for events such as this one: https://wayland.app/protocols/ext-transient-seat-v1#ext_transient_seat_v1:event:ready

### DIFF
--- a/wayland-backend/src/rs/server_impl/handle.rs
+++ b/wayland-backend/src/rs/server_impl/handle.rs
@@ -14,7 +14,7 @@ use crate::{
 
 use super::{
     client::ClientStore, registry::Registry, ClientData, ClientId, Credentials, GlobalHandler,
-    InnerClientId, InnerGlobalId, InnerObjectId, ObjectData, ObjectId,
+    GlobalId, InnerClientId, InnerGlobalId, InnerObjectId, ObjectData, ObjectId,
 };
 
 pub(crate) type PendingDestructor<D> = (Arc<dyn ObjectData<D>>, InnerClientId, InnerObjectId);
@@ -252,6 +252,10 @@ impl InnerHandle {
         self.state.lock().unwrap().global_info(id)
     }
 
+    pub fn global_name(&self, global: InnerGlobalId, client: InnerClientId) -> Option<u32> {
+        self.state.lock().unwrap().global_name(global, client)
+    }
+
     pub fn get_global_handler<D: 'static>(
         &self,
         id: InnerGlobalId,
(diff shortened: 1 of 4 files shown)


elinorbgr:
Given the CI failures, I think this bumps the minimum supported libwayland version. That is not a problem, but the CI scripts should be updated as such, and it should be documented in the changelogs.

elinorbgr:
Ok, actually the CI no longer fixes the libwayland version, so there's nothing to update here.

But it looks like this change requires a newer version of libwayland-server than is available on the CI environment...

PolyMeilex:
@elinorbgr I bumped Ubuntu version to 24.04, and we are green now

elinorbgr (approved):

ids1024:
So outside of the CI coverage tests (which were failing) this function is only needed if it's used, and won't break `server_system` on Ubuntu 20.04, Debian Stable, etc. that have a version of libwayland older than 1.22?

ids1024:
No, it looks like with this anvil fails to link on 22.04 with `wayland-backend/server_system` but without `wayland-sys/dlopen`. And fails at runtime if `wayland-sys/dlopen` is used. Whether or not `global_name` is ever used.

This seems potentially problematic? I think this will break 22.04 builds of cosmic-panel (which if I recall current is using `server_system`, while cosmic-comp isn't).

elinorbgr:
Could maybe this functionality be hidden behind a cargo feature?

PolyMeilex:
How about we go libinput-rs or gtk-rs like route, and have `libwayland-1.x` feature for each new version from now on?

ids1024:
I've just done a wayland-rs release without this for now. Presumably no one needs this functionality in their compositor immediately and it can wait for the next wayland-rs release (but otherwise we can just do another release when we figure out conditional compilation here).

Using a feature naming scheme like libinput-rs or gtk-rs sounds good to me.

I guess `external_library!` would need to be changed to allow attributes and then include them in the output of the macro? Seems a bit annoying to deal with, but I'm not sure how else it could be done.

PolyMeilex:
> I've just done a wayland-rs release without this for now.

That's totally fine, it's more of a future proofing for people that might need `ext_transient_seat_v1` in the future. I don't care if it lands today or in a year :smile:

ids1024:
Right.

Protocols like https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/40 may also want to make use of this. But it doesn't currently. (And "in a year" may be an optimistic timeline for getting things like that merged upstream...)

ids1024:
With https://github.com/elinorbgr/dlib/pull/22, it should be possible to just use:

```rust
#[cfg(feature = "libwayland_1_22")]
fn wl_global_get_name(*mut wl_global, *mut wl_client) -> u32,
```

ids1024:
Merged in https://github.com/Smithay/wayland-rs/pull/873 behind a feature flag.

Issue #728, opened 2024-05-16
Apparent misuse of scoped-tls, only appearing on a Raspberry Pi


KnorrFG:
Hi, I'm trying to run a GUI application (using Iced) on a Raspberry Pi 4B with the latest RPI Os. I didn't modify it yet.

The program crashes before any gui is created, with the following backtrace:

```
thread 'main' panicked at /home/neuro/.cargo/registry/src/index.crates.io-6f17d22bba15001f/scoped-tls-1.0.1/src/lib.rs:168:9:
cannot access a scoped thread local variable without calling `set` first
stack backtrace:
   0: std::panicking::begin_panic
   1: scoped_tls::ScopedKey<T>::with
   2: wayland_backend::sys::client_impl::dispatcher_func
   3: <unknown>
   4: <unknown>
   5: wl_display_dispatch_queue_pending
   6: wl_display_roundtrip_queue
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: wgpu_hal::vulkan::instance::<impl wgpu_hal::Surface<wgpu_hal::vulkan::Api> for wgpu_hal::vulkan::Surface>::configure
  12: wgpu_core::device::global::<impl wgpu_core::global::Global<G>>::surface_configure
  13: <wgpu::backend::wgpu_core::ContextWgpuCore as wgpu::context::Context>::surface_configure
  14: <T as wgpu::context::DynContext>::surface_configure
  15: wgpu::Surface::configure
  16: <iced_wgpu::window::compositor::Compositor as iced_graphics::compositor::Compositor>::configure_surface
  17: <iced_wgpu::window::compositor::Compositor as iced_graphics::compositor::Compositor>::create_surface
  18: <iced_renderer::compositor::Compositor as iced_graphics::compositor::Compositor>::create_surface
  19: core::ops::function::impls::<impl core::ops::function::FnMut<A> for &mut F>::call_mut
  20: winit::platform_impl::platform::wayland::event_loop::EventLoop<T>::pump_events
  21: winit::platform_impl::platform::wayland::event_loop::EventLoop<T>::run_on_demand
  22: iced_winit::application::run
  23: ts_updater::gui::gui_main
  24: ts_updater::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

The crash originates in scoped-tls, but it seems more like a bug in `wayland_backend`.
[...]

PR #738, opened 2024-06-20
backend: use distinct event queue for client sys

cmeissl:
some drivers seem to unconditionally dispatch
the default event queue. this can lead to spurious dispatching of wayland-rs managed objects breaking the tls invariant.
prevent this from happening by moving everything
off the default event queue to a distinct event queue

### DIFF
--- a/wayland-backend/src/sys/client_impl/mod.rs
+++ b/wayland-backend/src/sys/client_impl/mod.rs
@@ -168,6 +168,7 @@ struct ProxyUserData {
 #[derive(Debug)]
 struct ConnectionState {
     display: *mut wl_display,
+    owns_display: bool,
     evq: *mut wl_event_queue,
     display_id: InnerObjectId,
     last_error: Option<WaylandError>,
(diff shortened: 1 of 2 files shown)


cmeissl:
might also fix #728

wayland-backend/src/sys/client_impl/mod.rs
```diff
@@ -648,7 +624,7 @@ impl InnerBackend {
             }
         }
 
-        let ret = if guard.evq.is_null() || child_spec.is_none() {
+        let ret = if child_spec.is_none() {
             unsafe {
                 ffi_dispatch!(
                     wayland_client_handle(),
```
elinorbgr, line 630: While this change is technically correct, given the new semantics it would be better to change the "if" branch to invoke `wl_proxy_marshal_array()` instead of `wl_proxy_marshal_array_constructor_versioned()`, as this branch no longer creates new objects ever. That would make this code less confusing to read.

Switching to `wl_proxy_marshal_array()` also requires removing the last two arguments of the call.

cmeissl: makes sense, thanks! changed to use `wl_proxy_marshal_array` for the child_spec none branch.

elinorbgr (approved):
LGTM

PR #741, opened 2024-06-25
backend: delete dead clients from epoll

Drakulix:
I am not sure, if we need to do the same thing here for kqueue, I only observed this problem on an epoll system, but I figured it can't hurt.

This fixes a deadlock from killed clients blocking the loop in `dispatch_all_clients` for the rust-backend.

### DIFF
--- a/wayland-backend/src/rs/server_impl/common_poll.rs
+++ b/wayland-backend/src/rs/server_impl/common_poll.rs
@@ -160,7 +160,37 @@ impl<D> InnerBackend<D> {
                                 return Err(e);
                             }
                         }
-                        Err(e) => return Err(e),
+                        Err(e) => {
+                            #[cfg(any(target_os = "linux", target_os = "android"))]
+                            {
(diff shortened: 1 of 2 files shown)


wayland-backend/src/rs/server_impl/common_poll.rs
```diff
@@ -142,6 +142,7 @@ impl<D> InnerBackend<D> {
     pub(crate) fn dispatch_events_for(
         &self,
         data: &mut D,
+        poll_fd: BorrowedFd<'_>,
```
elinorbgr, line 145: You don't need to add this argument here, the poll fd can already be accessed in `state.poll_fd`

elinorbgr (changes requested):

wayland-backend/src/rs/server_impl/common_poll.rs
Drakulix: Ah, I read over the state lock. Fixed!

elinorbgr (approved):

Issue #747, opened 2024-07-22
Regression in foreign object management


chrisduerr:
In 41bab578c759065b9be95ccc952ed73557a042ab, support for managing foreign objects was added to wayland-backend, however before a new release was made it was already broken again in e777e85f6846ee732ab33adce5755bfc57b08649.

I'm not too familiar with the motivation behind e777e85f6846ee732ab33adce5755bfc57b08649 (cc @cmeissl), but my testing did show that despite compiling just fine, my application is leaking buffers again.

chrisduerr:
This was fixed by https://github.com/Smithay/wayland-rs/pull/750.

PR #750, opened 2024-07-27
backend/sys: associate imported proxy with correct event queue

cmeissl:
fixes a regression introduced in e777e85f6846ee732ab33adce5755bfc57b08649 which moved all proxies to a separate event queue but left proxies imported with `Backend::manage_object` on the default queue.

cc @chrisduerr

### DIFF
--- a/wayland-backend/src/sys/client_impl/mod.rs
+++ b/wayland-backend/src/sys/client_impl/mod.rs
@@ -757,6 +757,11 @@ impl InnerBackend {
     }
 
     /// Start managing a Wayland object.
+    ///
+    /// Safety: This will change the event queue the proxy is associated with.
+    /// Changing the event queue of an existing proxy is not thread-safe.
+    /// If another thread is concurrently reading the wayland socket and the
+    /// proxy already received an event it might get enqueued on the old event queue.
     pub unsafe fn manage_object(
         &self,
         interface: &'static Interface,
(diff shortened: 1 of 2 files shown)


wayland-backend/src/sys/client_impl/mod.rs
```diff
@@ -764,7 +764,10 @@ impl InnerBackend {
         data: Arc<dyn ObjectData>,
     ) -> ObjectId {
         let mut guard = self.lock_state();
-        unsafe { self.manage_object_internal(interface, proxy, data, &mut guard) }
+        unsafe {
+            ffi_dispatch!(wayland_client_handle(), wl_proxy_set_queue, proxy, guard.evq);
```
elinorbgr, line 768: One thing to keep in mind is that *in theory* this is racy: changing the event queue of an existing proxy is not thread-safe (which is why the whole "proxy wrapper" stuff exists when creating proxies: so that they are lerictly created on the correct queue).

I think in most cases this is fine, but I think this should still be documented in the public API docs.

The core of the risk, is that if another thread is concurrently reading from the socket while this is being invoked, and the object that is in the process of being managed already received an event, that event may end up enqueued on the old queue, which may in turn cause dispatching issues or unsafety.

This is really a corner-case situation, and mostly of relevance for objects that are susceptible to receive events as soon as they are created, but I think this would be better written down so as to not surprise anyone.

elinorbgr (changes requested):

chrisduerr:
I can confirm that this patch fixes the leak.

chrisduerr:
Any reason to hold off on this? Seems better than the status quo.

Also wanted to link #747, so that gets closed when this is merged.

wayland-backend/src/sys/client_impl/mod.rs
cmeissl: Added a safety warning, is this good enough?

elinorbgr (approved):

ids1024 (approved):

PR #752, opened 2024-08-10
Allow DOCTYPE in protocol definition

Nxllpointer:
Currently wayland-scanner only allows the `<?xml ... ?>` declaration and the protocol tag at the start of a xml protocol definition. This PR allows one to additionaly specify a `<!DOCTYPE ...>` like this:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE protocol SYSTEM "https://gitlab.freedesktop.org/wayland/wayland/-/raw/ee12e69b8fb8c55460126e4e527eb4f7fee5526e/protocol/wayland.dtd">
<protocol name="test_proto">
    <interface name="test_interface" version="1">
        <description summary="Testing"></description>
        <event name="ping"></event>
    </interface>
</protocol>
```
The specified file can be consumed by LSPs to enable assisting with autocompletion and syntax checking.

### DIFF
--- a/wayland-scanner/src/parse.rs
+++ b/wayland-scanner/src/parse.rs
@@ -54,9 +54,9 @@ fn parse_or_panic<T: FromStr>(txt: &[u8]) -> T {
 
 fn init_protocol<R: BufRead>(reader: &mut Reader<R>) -> Protocol {
     // Check two firsts lines for protocol tag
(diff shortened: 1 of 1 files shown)


ids1024 (approved):

Nxllpointer:
Thank you

PR #765, opened 2024-10-23
chore: Release

ids1024:
`cargo release --no-publish --no-tag --no-push --execute patch --prev-tag-name release-2024-09-04 --exclude wayland-sys --exclude wayland-backend --exclude wayland-scanner --exclude wayland-egl`

If on `master` with `prev-tag-name` specified, the unchanged crates produce warnings, but are not automatically excluded. Not sure if there's some way to do that.

### DIFF
--- a/wayland-client/CHANGELOG.md
+++ b/wayland-client/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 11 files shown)


ids1024:
I guess https://github.com/crate-ci/cargo-release/issues/298 would be helpful.

PR #769, opened 2024-11-27
Fix docs typo in event_created_child

SUPERCILEX:


wayland-client/src/event_queue.rs
```diff
@@ -140,7 +140,7 @@ where
 /// initialization of the user data for event-created objects. The usage syntax is as follow:
 ///
 /// ```ignore
-/// impl Dispatch<WlFoo, FooUserData> for MyState {
+/// impl Dispatch<MyState, FooUserData> for MyState {
```
elinorbgr, line 143: No, this is not a typo, the type argument here should be `WlFoo`, as a generic stand-in for the Wayland interface for which `Dispatch` is being implemented.

SUPERCILEX:
Oh shoot you're right, my bad.

PR #772, opened 2024-12-14
backend/rs: delay client cleanup

cmeissl:
delay client cleanup until the state lock is dropped to prevent potential deadlocks in drop impls of the client state.


Drakulix (approved):

ids1024 (approved):
Looks like a reasonable way to avoid the deadlock here.

PR #773, opened 2024-12-15
Remove a pointless allocation

SUPERCILEX:
The stack size is 152 bytes which is teeny.

### DIFF
--- a/wayland-backend/Cargo.toml
+++ b/wayland-backend/Cargo.toml
@@ -31,7 +31,7 @@ features = [
 ]
 
 [dependencies.rustix]
-version = "0.38.17"
+version = "0.38.42"
 features = [
     "event",
     "fs",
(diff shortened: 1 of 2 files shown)


SUPERCILEX:
None of the failures look related.

ids1024:
Yeah, looks like https://github.com/Smithay/wayland-rs/pull/772 has fixes for the CI errors here.

I thought there was a reason this used a vec... and it looks like I was right, but this has been fixed in Rustix: https://github.com/bytecodealliance/rustix/pull/889. (Which I now remember seeing.)

I guess we should update the `rustix` dependency to explictly require  0.38.21 or newer? (I think this would be failing in the `check-minimal` CI test because of that; if it wasn't failing earlier for unrelated reasons.)

SUPERCILEX:
Bumping rustix sounds good and I'll wait for the other PR.

ids1024:
That PR is merged.

If CI passes after rebasing and updating the `rustix` dep version, this looks good.

SUPERCILEX:
All green

ids1024 (approved):

ids1024:
Thanks!

PR #777, opened 2025-01-13
Update downcast-rs to 2.0

torokati44:
Simple housekeeping, no other particular reason.

This would deduplicate in the future what https://github.com/ruffle-rs/ruffle/pull/19205/files duplicates for us.


torokati44:
See the changelog: https://github.com/marcianx/downcast-rs/blob/master/CHANGELOG.md#200---2025-01-06

ids1024:
https://docs.rs/wayland-backend/0.3.7/wayland_backend/sys/client/struct.Backend.html#impl-Downcast-for-T

It seems the public API of `wayland-backend` does export types implementing traits from downcast-rs, so this is (at least) technically a breaking API change.

ids1024:
`downcast-rs` dependency removed in https://github.com/Smithay/wayland-rs/pull/912.

Issue #782, opened 2025-02-04
wayland-backend test fails to build since #731


jamessan:
Since #731, in particular 06703518717675293228e7f9b0d60360723dd8ce, `cargo test -p wayland-backend --features server_system` fails to build:

```
error[E0433]: failed to resolve: could not find `client` in `sys`
   --> wayland-backend/tests/rs_sys_impls.rs:155:17
    |
155 |             dyn client::ObjectData: std::fmt::Debug, downcast_rs::DowncastSync
    |                 ^^^^^^ could not find `client` in `sys`
    |
note: found an item that was configured out
   --> /home/jamessan/src/github.com/wayland-rs/wayland-backend/src/sys/mod.rs:39:9
    |
39  | pub mod client;
    |         ^^^^^^
note: the item is gated here
   --> /home/jamessan/src/github.com/wayland-rs/wayland-backend/src/sys/mod.rs:37:1
    |
37  | #[cfg(any(test, feature = "client_system"))]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: consider importing this module
    |
151 +     use wayland_backend::client;
    |

error[E0433]: failed to resolve: could not find `client` in `sys`
   --> wayland-backend/tests/rs_sys_impls.rs:160:13
    |
160 |             client::ObjectId: std::fmt::Debug,
    |             ^^^^^^ could not find `client` in `sys`
    |
note: found an item that was configured out
   --> /home/jamessan/src/github.com/wayland-rs/wayland-backend/src/sys/mod.rs:39:9
    |
39  | pub mod client;
    |         ^^^^^^
note: the item is gated here
   --> /home/jamessan/src/github.com/wayland-rs/wayland-backend/src/sys/mod.rs:37:1
    |
37  | #[cfg(any(test, feature = "client_system"))]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: consider importing this module
    |
151 +     use wayland_backend::client;
    |

error[E0433]: failed to resolve: could not find `client` in `sys`
   --> wayland-backend/tests/rs_sys_impls.rs:170:22
    |
170 |         assert_impl!(client::Backend: Send, Sync, std::fmt::Debug);
    |                      ^^^^^^ could not find `client` in `sys`
    |
note: found an item that was configured out
[...]

PR #783, opened 2025-02-06
backend: Add scoped-tls as a dev-dependency

jamessan:
Since aa8607e moved scoped-tls to an optional dependency,
`cargo test -p wayland-backend` fails to build (independent from #782)
since neither the client_system nor server_system features are enabled:

    error[E0432]: unresolved import `scoped_tls`
      --> wayland-backend/src/sys/client_impl/mod.rs:26:5
       |
    26 | use scoped_tls::scoped_thread_local;
       |     ^^^^^^^^^^ use of undeclared crate or module `scoped_tls`

    error[E0432]: unresolved import `scoped_tls`
      --> wayland-backend/src/sys/server_impl/mod.rs:20:5
       |
    20 | use scoped_tls::scoped_thread_local;
       |     ^^^^^^^^^^ use of undeclared crate or module `scoped_tls`

    error[E0425]: cannot find value `BACKEND` in this scope
       --> wayland-backend/src/sys/client_impl/mod.rs:378:19
        |
    378 |         let ret = BACKEND.set(&backend, || unsafe {
        |                   ^^^^^^^ help: a local variable with a similar name exists: `backend`

There is code gated with `#[cfg(any(test, feature = "..._system"))]`
which therefore requires scoped-tls during test.

### DIFF
--- a/wayland-backend/Cargo.toml
+++ b/wayland-backend/Cargo.toml
@@ -47,6 +47,7 @@ wayland-scanner = { path = "../wayland-scanner"}
 wayland-sys = { path = "../wayland-sys", features = ["client", "server"] }
 concat-idents = "1.1"
 env_logger = "0.10"
+scoped-tls = "1.0"
 
 [features]
 client_system = ["wayland-sys/client", "dep:scoped-tls"]


ids1024 (approved):

ids1024:
Ah yeah, I didn't consider the `any(test, ...` in the feature gate when changing this. Thanks!

PR #790, opened 2025-02-19
protocols-plasma: Update to `plasma-wayland-protocols` 1.16.0

ids1024:
Some protocols have [added a warning](https://invent.kde.org/libraries/plasma-wayland-protocols/-/merge_requests/80):

```
Warning! The protocol described in this file is a desktop environment
implementation detail. Regular clients must not use this protocol.
Backward incompatible changes may be added without bumping the major
version of the extension.
```

While protocols like `blur` presumably are considered safe to use.

Perhaps we should have a feature flag for those "implementation detail" protocols. But for now, this is just an update without breaking anything.

### DIFF
--- a/wayland-protocols-plasma/CHANGELOG.md
+++ b/wayland-protocols-plasma/CHANGELOG.md
@@ -13,6 +13,9 @@
   * `keystate` version 5
   * `plasma-window-management` version 18
   * `zkde-screencast-unstable-v1` version 4
+- Add bindings that were missing
+  * `kde-lockscreen-overlay-v1`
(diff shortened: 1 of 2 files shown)


fnr1r:
Small note: CI failed because  `plasma-wayland-protocols/src/protocols/plasma-virtual-desktop.xml` was renamed to `plasma-wayland-protocols/src/protocols/org-kde-plasma-virtual-desktop.xml`

ids1024:
Yep. I also forgot to add the new protocols to `lib.rs`.

Issue #596, opened 2023-01-04
`destroy()` Apis should be called on Drop


A6GibKm:
When the destroy() api's sole purpose is to delete the object on the C side, the api should be `fn destroy(self)` rather than `pub fn destroy(&self)` and be called automatically on Drop.

i509VCB:
I think this is more appropriately addressed in the implementing libraries like sctk and smithay.

A6GibKm:
I mean the objects are completely useless after calling destroy, `destroy` should totally not take a reference and leave you with a useless object.

cmeissl:
At least on the server a destroyed surface for example can still be usable. Smithay puts all the surface state including the imported textures in the surface user data. This could be used for fade-out animations to give an example. Not sure if there are use-cases for that in a client.

ids1024:
One solution could be to destroy on drop by default, but method to set a flag on an object to disable automatic calls to the destructor. That might be more intuitive since Rust users expect a "destructor" to be called automatically through RAII, and may help make sure objects aren't accidentally leaked. Though that's more complicated in some ways.

Requiring users of the library to keep a reference to any object they don't want destroyed seems problematic though.

Otherwise, perhaps documentation can be improved. It's certainly surprising behavior, even if it ultimately makes sense.

axelkar:
Could destroy be made into a trait so I can make a "destroy on drop" wrapper easily?

SUPERCILEX:
I would love a trait as I just implemented that idea for myself:
```rust
trait Destroyable {
    fn destroy(&self);
}

impl Destroyable for WlSeat {
    fn destroy(&self) {
        self.release();
    }
}

impl Destroyable for ZwlrDataControlManagerV1 {
    fn destroy(&self) {
        self.destroy();
    }
}

impl Destroyable for ZwlrDataControlDeviceV1 {
    fn destroy(&self) {
        self.destroy();
    }
}

impl Destroyable for ZwlrDataControlOfferV1 {
    fn destroy(&self) {
        self.destroy();
    }
}

struct AutoDestroy<T: Destroyable>(T);

impl<T: Destroyable> AutoDestroy<T> {
    fn into_inner(self) -> T {
        let this = ManuallyDrop::new(self);
        unsafe { ptr::read(&this.0) }
    }
}

impl<T: Destroyable + Debug> Debug for AutoDestroy<T> {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
[...]

PR #792, opened 2025-02-27
WIP client: Add a `DestroyOnDrop` wrapper that calls destructor on `Drop`

ids1024:
I've been thinking a helper like this could be handy, since I end up implementing `Drop` on types that contain proxies often. So this would save some code, and also allow destructuring of those wrapper types since they won't have their own `Drop` impls.

The same idea was also suggested in https://github.com/Smithay/wayland-rs/issues/596.

I need to test this in a few use cases, and check a few things (probably should test the object version; make sure wayland doesn't allow multiple destructors or destructors with arguments, or handle those somehow; compare the `_destroy` code generated in libwayland.)

### DIFF
--- a/wayland-client/src/lib.rs
+++ b/wayland-client/src/lib.rs
@@ -321,6 +321,31 @@ pub trait Proxy: Clone + std::fmt::Debug + Sized {
     }
 }
 
+/// Trait for instances of [`Proxy`] that have destructors
+pub trait ProxyDestroy: Proxy {
+    /// Sends the destructor request to the proxy, which is usually called `.destroy()`
+    fn call_proxy_destructor(&self);
+}
+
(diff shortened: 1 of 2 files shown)


ids1024:
Looks like this is failing to generate bindings for https://wayland.app/protocols/color-management-v1 since it has objects with "destructor" requests called `::create` that take a `new_id` argument, rather than no arguments like normal destructors.

(This is in contrast to the design of https://wayland.app/protocols/linux-dmabuf-v1, where `destroy` is actually a separate request that has to be sent after `create`/`create_immed`.)

ids1024:
We wouldn't want to restrict it to only destructors called `::destroy`, since it should probably work with globals that have a `::release` destructor.

Could restrict it to just destructors that take no arguments.

Issue #791, opened 2025-02-21
Don't use `WEnum` for bitflags?


ids1024:
If I recall, bitflags 1.x was different, but bitflags 2.x allows constructing a flags instance with unrecognized fields: https://docs.rs/bitflags/latest/bitflags/trait.Flags.html#tymethod.from_bits_retain.

This way it is easy to test for recognized bitflags, but still easy to see what unknown flags are set if needed. Probably better not to wrap this in `WEnum`? Currently if any unrecognized bitflag is set, even if there are also recognized flags, it ends up a `WEnum::Unknown`.

See https://github.com/pop-os/cosmic-protocols/pull/51/commits/b1fa8c4df7b0f77eae5f77f7eb157df08eb55ed6 for an example of code that maps `WEnum` to the underlying bitfield, with unrecognized fields if any retained. So this is possible to wrap currently, but may be cleaner if that's just what wayland-rs did already.

See also https://github.com/Smithay/wayland-rs/issues/764 and https://gitlab.freedesktop.org/wayland/wayland/-/issues/497 for discussion around "open" and "closed" enums (that's primarily about non-bitflag enums).

PR #793, opened 2025-02-27
WIP Don't use `WEnum` for bitflags

ids1024:
https://github.com/Smithay/wayland-rs/issues/791.

If no one sees issues with this approach, it can be included whenever the next breaking release is.


ids1024:
Closing in favor of https://github.com/Smithay/wayland-rs/pull/908, which also changes how non-bitfiled enums work.

PR #794, opened 2025-03-11
Update to `rustix` 1.0

ids1024:


plugwash:
My searching indicates that the wayland-backend crate is likely affected by "silent behaviour changes" in rustix's event APIs.


> When a &mut Vec<_> is passed to [rustix::event::epoll::wait](https://docs.rs/rustix/1/rustix/event/epoll/fn.wait.html), [rustix::event::kqueue::kevent](https://docs.rs/rustix/1/x86_64-unknown-freebsd/rustix/event/kqueue/fn.kevent.html), or [rustix::event::port::getn](https://docs.rs/rustix/1/x86_64-unknown-illumos/rustix/event/port/fn.getn.html), these functions previously adjusted the length of the Vec to the number of elements written, and now do not. A common alternative is to wrap the &mut Vec<_> using [spare_capacity](https://docs.rs/rustix/1/rustix/buffer/fn.spare_capacity.html), and then to clear the Vec by iterating using .drain(..) after each call. For an example of using spare_capacity in this way, see [here](https://docs.rs/rustix/1/rustix/event/epoll/index.html#examples).

ids1024:
Okay, now I've taken the time to test locally on FreeBSD, I've updated it so the CI tests are passing on Linux and FreeBSD. I'll probably do a little more testing before merging it, but I think things should be good now.

PR #797, opened 2025-03-15
backend: hold the dispatch lock for send_request

kchibisov:
Don't allow calling `send_request` while we have dispatch_pending, since `send_request` can call `wl_proxy_add_dispatcher`, which requires user to synchronize access to it.

Otherwise the proxy might be created by `send_request` but then get a `null` deref if dispatch is called at the right time because the proxy data was not set.


kchibisov:
The BSD failure looks arbitrary.

PR #799, opened 2025-03-26
BREAKING: use `Cow<'static, CStr>` for string argument

### LINKED ISSUE #748: String arguments should be CString
If you pass `XdgToplevel::set_title` a String with embedded 0 bytes (which is valid for a `String`), you will get a panic with a backtrace that kinda looks like this:
```
thread 'main' panicked at ...
xdg.rs:138:5:
called `Result::unwrap()` on an `Err` value: NulError(<bytes>)
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: <wayland_protocols::xdg::shell::generated::client::xdg_toplevel::XdgToplevel as wayland_client::Proxy>::write_request
   4: wayland_client::conn::Connection::send_request
   5: wayland_protocols::xdg::shell::generated::client::xdg_toplevel::XdgToplevel::set_title
```
This is a fairly obtuse error, but digging into the library, it appears this is because string arguments are converted to CStrings internally, with an unwrap:
https://github.com/Smithay/wayland-rs/blob/df56d7d52ead01d8f65af437e793b5c3969f3958/wayland-scanner/src/common.rs#L541
If wayland-scanner doesn't want to handle this error, then it makes no sense for String arguments to be String, when they should instead just be `std::ffi::CString` and leave it on callers to deal with constructing it.

CrazyboyQCD:
Closes https://github.com/Smithay/wayland-rs/issues/748.
Leave construction of string argument to caller and make `ffi` process less panic-proned, and is benefit for users whose rust version is above 1.77 to use [C-string literals](https://doc.rust-lang.org/edition-guide/rust-2021/c-string-literals.html#c-string-literals)

### DIFF
--- a/wayland-backend/src/protocol.rs
+++ b/wayland-backend/src/protocol.rs
@@ -1,6 +1,6 @@
 //! Types and utilities for manipulating the Wayland protocol
 
-use std::{ffi::CString, os::unix::io::AsRawFd};
+use std::{borrow::Cow, ffi::CStr, os::unix::io::AsRawFd};
 
 pub use wayland_sys::common::{wl_argument, wl_interface, wl_message};
 
@@ -22,7 +22,7 @@ pub enum ArgumentType {
     Uint,
     /// A signed fixed point number with 1/256 precision
     Fixed,
-    /// A string. This is represented as a [`CString`] in a message.
+    /// A string. This is represented as a [`Cow<'static, CStr>`] in a message.
     Str(AllowNull),
     /// Id of a wayland object
     Object(AllowNull),
@@ -53,11 +53,11 @@ pub enum Argument<Id, Fd> {
     Uint(u32),
     /// A signed fixed point number with 1/256 precision
     Fixed(i32),
-    /// CString
+    /// `Cow<'static, CStr>`
     ///
     /// The value is boxed to reduce the stack size of Argument. The performance
     /// impact is negligible as `string` arguments are pretty rare in the protocol.
-    Str(Option<Box<CString>>),
+    Str(Option<Box<Cow<'static, CStr>>>),
     /// Id of a wayland object
     Object(Id),
     /// Id of a newly created wayland object
--- a/wayland-backend/src/rs/server_impl/client.rs
+++ b/wayland-backend/src/rs/server_impl/client.rs
@@ -282,7 +282,7 @@ impl<D> Client<D> {
                 [
                     Argument::Object(ObjectId { id: object_id.clone() }),
                     Argument::Uint(error_code),
-                    Argument::Str(Some(Box::new(message))),
+                    Argument::Str(Some(Box::new(message.into()))),
                 ],
             ),
             // wl_display.error is not a destructor, this argument will not be used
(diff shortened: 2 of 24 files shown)


ids1024:
If arguments take `Cow`, then all callers will have to explicitly convert to a `Cow`. A generic requiring `Into<Cow<'_, CStr>>` could be better, though that still isn't implemented for types like `&str`.

Perhaps a custom trait like Rustix uses would be best here: https://docs.rs/rustix/1.0.3/rustix/path/trait.Arg.html.

ids1024:
In principle using `CStr` in the API is quite correct since Wayland requires null-terminated strings, but it may be annoying to deal with.

i509VCB:
There is also the issue that strings in Wayland are rather undefined in their use. Nothing stops a client from sending you malformed UTF-8 or UTF-16.

The two exceptions are xdg-toplevel titles (explicitly defined as UTF-8) and the identifier the compositor created for ext-foreign-toplevel-handle (printable ASCII)

ids1024:
Right. As defined in https://wayland.freedesktop.org/docs/html/ch04.html, I guess technically the only difference between strings and arrays is null-termination? And this seems to imply there shouldn't be other nul bytes, though I don't see that explicitly mentioned.

So `CStr`/`CString` is correct here, though a pain to work with. Life would be easier if wayland was more specific and libwayland did utf-8 validation for strings.

CrazyboyQCD:
@ids1024 
After overview, I found a auto-convert trait argument will increase code complexity and make product bloaty, thought callers need to deal with this, it is more flexible for callers if they want to be more performant.

ids1024:
Yeah, there's not a clear best way to do this.

Converting a `&str` to a `CStr` manually is a bit annoying, but strings aren't really that common in wayland protocols (and higher level APIs like smithay and smithay-client-toolkit can decide what is suitable as a higher level API for a specific protocol). And having to deal with `CString`s that may or may not be valid UTF-8 in return values is necessary regardless.

orowith2os:
I will also note that the Wayland protocol defines a string argument as:

> Starts with an unsigned 32-bit length (including null terminator), followed by the string contents, including terminating null byte, then padding to a 32-bit boundary. A null value is represented with a length of 0. 

We need the null terminator, but it's not necessarily used for counting the string's length. That's something C-based Wayland clients can do. 

And here's a note in CStr's documentation on count_bytes:

> Note: This method is currently implemented as a constant-time cast, but it is planned to alter its definition in the future to perform the length calculation whenever this method is called.

It would be better to take a `&mut String` (or a `Cow<_, String>`) and check for a NUL byte, add it if one doesn't exist, and *then* pass it as an argument. It would also need to make sure that there's no NUL byte in the middle of the string.

orowith2os:
Recent versions of the core Wayland protocol will also require that all strings used are valid UTF-8: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/387

orowith2os:
Correction: Cow<str> would probably be better? Then append NUL to that. Would also need to perform validation so that it doesn't pass a string with two NUL bytes.

PR #802, opened 2025-04-06
backend/client_sys: add read_without_dispatch

kchibisov:
Add a way to read the event queue, but don't do an immediate dispatch
of the events, leaving it up to dispatch_inner_queue. This is useful,
when we spawn a separate thread that interacts with the main queue of
the backend, like for example to track if events generally arrived and
we have something to dispatch.

--

The use of such API is present here https://github.com/rust-windowing/winit/pull/4185 . We can not fix the issue without it in winit.


kchibisov:
ping @ids1024

PR #804, opened 2025-04-15
Update delegate_dispatch docs

dcz-self:
This fixes a typo and includes a description of the `@<>` syntax which allows delegatng over a generic type.

I needed this feature but it wasn't documented, so I was lucky to catch a dev and asked. With this description, even unlucky people can find this feature.


ids1024 (approved):
Seems like a good description and example of the `@` syntax.

Thanks!

Issue #806, opened 2025-04-24
wayland-backend: rust backend fails to limit the number of file descriptors sent with a single message to MAX_FDS_OUT = 28 causing problems


gergo-salyi:
### Symptoms

Using this wayland-rs crate in a client program using the `rust backend`, interacting with a libwayland server (e.g. Sway), if **between two** `wayland_client::Connection::flush` I try to do:
- more than 28 `smithay_client_toolkit::shm::raw::RawPool::new`, it will result in a protocol error on next read: `Protocol error 1 on object wl_display@1: invalid arguments for wl_shm#5.create_pool`
- more than 28 `smithay_client_toolkit::dmabuf::DmabufParams::add`, it will result in a protocol error on next read: `Protocol error 1 on object wl_display@1: invalid arguments for zwp_linux_buffer_params_v1#28.add`

These symptoms do not happen when using the `sys backend` (`client_system` feature enabled), then it works fine.

### Cause

`wayland_backend::rs::socket::Socket::send_msg` has a comment on top that it should not send more than MAX_FDS_OUT = 28 file descriptors with a single message. However the function itself if does not enforce this limit.

https://github.com/Smithay/wayland-rs/blob/cd317b694a392c50a9f4d6cfe14f772423d48c4d/wayland-backend/src/rs/socket.rs#L39-L42

When the higher level interface calls `wayland_client::Connection::flush`, the rust backend will get here through `wayland_backend::rs::socket::BufferedSocket::flush` which fails to enforce this limit too and calls `send_msg` with an arbitrary number of outgoing file descriptors in its ring buffer:

https://github.com/Smithay/wayland-rs/blob/cd317b694a392c50a9f4d6cfe14f772423d48c4d/wayland-backend/src/rs/socket.rs#L150-L157

Then the Wayland peer will lose the excess fds in recvmsg after MAX_FDS_OUT because it only provides buffer for MAX_FDS_OUT size of ancillary data, then eventually produce an obscure error when it tries to access the missing fd.

### Proposed Solution

More than MAX_FDS_OUT = 28 file descriptors must be sent with multiple sendmsg syscalls.

Wayland-backend's rust backend should either
1) try to prevent the outgoing fd buffer to grow above MAX_FDS_OUT and implicitly flush when queuing more fds than that,
2) allow the buffer to accumulate more than MAX_FDS_OUT fds but split that flush across multiple sendmsg syscalls

[...]

PR #812, opened 2025-04-29
backend/rs: enforce MAX_FDS_OUT limit

gergo-salyi:
backend/rs: do flush as multiple sendmsg if needed to enforce MAX_FDS_OUT limit

Fixes https://github.com/Smithay/wayland-rs/issues/806

### DIFF
--- a/wayland-backend/src/rs/socket.rs
+++ b/wayland-backend/src/rs/socket.rs
@@ -148,17 +148,52 @@ impl BufferedSocket {
 
     /// Flush the contents of the outgoing buffer into the socket
     pub fn flush(&mut self) -> IoResult<()> {
-        let written = {
-            let bytes = self.out_data.get_contents();
-            if bytes.is_empty() {
-                return Ok(());
+        let bytes = self.out_data.get_contents();
+        let fds = &self.out_fds;
+        let mut written_bytes = 0;
+        let mut written_fds = 0;
+        let mut ret = Ok(());
+        while written_bytes < bytes.len() {
+            let mut bytes_to_write = &bytes[written_bytes..];
+            let mut fds_to_write = &fds[written_fds..];
+            if fds_to_write.len() > MAX_FDS_OUT {
+                // While we need to send more than MAX_FDS_OUT fds,
+                // send them with separate send_msg calls in MAX_FDS_OUT sized chunks
+                // together with 1 byte of normal data.
+                // This achieves the same that libwayland does in wl_connection_flush
+                // and ensures that all file descriptors are sent
+                // before we run out of bytes of normal data to send.
+                bytes_to_write = &bytes_to_write[..1];
+                fds_to_write = &fds_to_write[..MAX_FDS_OUT];
             }
-            self.socket.send_msg(bytes, &self.out_fds)?
-        };
-        self.out_data.offset(written);
+            if bytes_to_write.len() > MAX_BYTES_OUT {
+                // Also ensure the MAX_BYTES_OUT limit, this stays redundant as long
+                // as self.out_data has a fixed MAX_BYTES_OUT size and cannot grow.
+                bytes_to_write = &bytes_to_write[..MAX_BYTES_OUT];
+            }
+            match self.socket.send_msg(bytes_to_write, fds_to_write) {
+                Ok(0) => {
+                    // This branch should be unreachable because
+                    // a non-zero sized send or sendmsg should never return 0
+                    written_fds += fds_to_write.len();
+                    break;
+                }
+                Ok(count) => {
+                    written_bytes += count;
+                    written_fds += fds_to_write.len();
+                }
+                Err(error) => {
+                    ret = Err(error);
+                    break;
+                }
+            }
+        }
+        // Either the flush is done or got an error, so clean up and
+        // remove the data and the fds which were sent form the outgoing buffers
+        self.out_data.offset(written_bytes);
         self.out_data.move_to_front();
-        self.out_fds.clear();
-        Ok(())
+        self.out_fds.drain(..written_fds);
+        ret
     }
 
     // internal method
--- a/wayland-tests/tests/backend_socket_out_limits.rs
+++ b/wayland-tests/tests/backend_socket_out_limits.rs
@@ -1,0 +1,243 @@
+extern crate tempfile;
+
+use std::io::ErrorKind;
+use std::os::unix::io::{AsFd, OwnedFd};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+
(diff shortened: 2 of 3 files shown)


gergo-salyi:
The idea to first send out fds together with 1 byte of normal data until we have fds <= MAX_FDS_OUT comes from libwayland, see:
https://gitlab.freedesktop.org/wayland/wayland/-/blob/6137c8c21383d7ebb92cf3331be56ca0aecccd5d/src/connection.c#L445

gergo-salyi:
That CI looks no good, my first guess is that test helper function `wayland_tests::helpers::roundtrip` wasn't meant to handle WouldBlock which happens when testing with the larger amount of data sent through the socket there. (I think I didn't see that problem on my machine because my kernel buffers are larger than the CI runner.)

I need to give my test a patched version of `wayland_tests::helpers::roundtrip` which handles WouldBlock, please wait for that.

gergo-salyi:
My bad. 

I applied that clippy lint to the PR. We can try again.

wayland-backend/src/rs/socket.rs
```diff
@@ -148,17 +148,46 @@ impl BufferedSocket {
...
+            if bytes_to_write.len() > MAX_BYTES_OUT {
+                // Also ensure the MAX_BYTES_OUT limit, this stays redundant as long
+                // as self.out_data has a fixed MAX_BYTES_OUT size and cannot grow.
+                bytes_to_write = &bytes_to_write[..MAX_BYTES_OUT];
+            }
+            match self.socket.send_msg(bytes_to_write, fds_to_write) {
+                Ok(count) => {
+                    written_bytes += count;
+                    written_fds += fds_to_write.len();
+                }
```
ids1024, line 178: Could `send_msg` return `Ok(0)` if the other end of the socket has been `close`d or `shutdown`? If so, this would loop infinitely.

I guess it should have a `Ok(0) => { break; }` branch.

wayland-tests/tests/backend_socket_out_limits.rs
```diff
@@ -0,0 +1,243 @@
...
+        .send_request(
+            &client.display,
+            wayc::protocol::wl_display::Request::Sync {},
+            Some(Arc::new(SyncData { done })),
+        )
+        .unwrap();
+    // Proper WouldBlock handling would need polling like in real applications.
+    // So just ignore WouldBlock up to 1000 times,
+    // that is enough for this test if the loop makes progress,
+    // and avoids the busy loop if the test is failing with no progress.
```
ids1024, line 174: So it is a busy loop, but limited to just 1000 iterations?

Probably not that much harder to call https://docs.rs/rustix/0.38.44/rustix/event/fn.poll.html with a timeout? I think it would work to just call `poll` at the start of each iteration and removing the counter here.

wayland-backend/src/rs/socket.rs
gergo-salyi: As far as I understand Unix systems and as far as I can read into `man 2 send`, the "return 0 on end-of-file" behavior is for `read` / `recv` / `recvmsg`.

The `write` / `send` / `sendmsg` syscalls do not "return 0 on end-of-file". They return -1 and `EPIPE`. This applies to all the cases where: our side of the socket is shutdown for write, or the peer's side is shutdown for read, or the peer closed the socket. _Please correct me if I'm wrong here._

The only way `send` or `sendmsg` returns 0 if I try to do a 0 sized send. Which never happens inside the loop because `while written_bytes < bytes.len()` ensures `bytes_to_write = &bytes[written_bytes..]` has a length larger than 0. (This is also why later `bytes_to_write = &bytes_to_write[..1]` is panic safe.)

So I think `Ok(0)` branch is unreachable unless:
 - someone not careful later changes `Socket::send_msg` and does a bad change e.g. maps EPIPE to Ok(0)
 - we are running on a non POSIX conforming kernel which returns 0 instead of EWOULDBLOCK or EPIPE.

(And btw what happened to `fds_to_write` in this case? I can only guess about that.)

But okay, better safe than sorry, I add an `Ok(0)` branch with the philosophy to keep the previous behavior which is:
```
Ok(0) => {
    // This branch should be unreachable because
    // a non-zero sized send or sendmsg should never return 0
    written_fds += fds_to_write.len();
    break;
}
```

wayland-tests/tests/backend_socket_out_limits.rs
gergo-salyi: Yes, a busy loop limited to 1000 WouldBlock or here ~250 iterations because we can get 4 WouldBlock per iteration~ (or more like 500 iterations because now looking at it I realized `server.display.dispatch_clients` and `server.display.flush_clients` currently do not return `WouldBlock` but they return `Ok` instead. Which does not change the correctness of the test).

I don't understand the timeout part of the proposed poll. `fn roundtrip` in the test is a single threaded, sequence in a loop: 
```
while !roundtrip_done {
    client.flush();
    server.read(); server.dispatch();
    server.flush();
    client.read(); client.dispatch();
}
```
If if I insert a `poll` anywhere it will enter and block inside a syscall and see the the readable or writable status of client or server socket. But that will not change over time, there is no other thread that could change that during the timeout.

And what should I poll for? Only because one socket is readable or writable at the beginning of each iteration that does not mean that the loop makes progress. Also, wayland-server does not give me a socket fd (?), but it gives me an epoll fd registered with all the clients (?). (I'm not really familiar with the server API.)

That busy loop with 250-1000 iterations is not nice but gets the job done for the test. I would rather not introduce a point of failure there with polling and/or with separate client/server threads. The current test works (fails without the MAX_FDS_OUT fix, and passes with the fix). If there is no correctness concern then  I think it should stay.

gergo-salyi:
Updated PR, changed the fix commit, did not change the test commit, see comments below the reviews

wayland-backend/src/rs/socket.rs
ids1024: Yeah, I guess the implication of `send(2)` and `sendmsg(2)` not mentioning a 0 return other than if the len argument is 0 (unlike `recv(2)` and `recvmsg(2)`) is that it won't otherwise return 0. And returns `EPIPE` where recv would return `0`.

wayland-tests/tests/backend_socket_out_limits.rs
ids1024: Oh yeah, it's handling the client and server in one loop. It might be a bit more complicated to make it correctly use polling, and it's probably fine for a test.

ids1024 (approved):

ids1024:
Thanks for working on this! It's good to have these things working correctly, even if it's fairly uncommon in practice.

Issue #717, opened 2024-04-26
Failed build when running with coverage on stable Rust


tgross35:
Both `wayland-backend` and `wayland-client` have a line like the following: https://github.com/Smithay/wayland-rs/blob/1fc7814bee5d90f83bab428c8167af43201f8a9b/wayland-backend/src/lib.rs#L46

This is a problem because if building a project with coverage will always try to enable this feature, which will fail if using a stable rustc. This should be changed so the `coverage_attribute` feature and `coverage(...)` attributes to go with it are not enabled unless compiled with nightly.

I think the easiest fix is probably to add a Cargo feature gate for nightly features and use it in place of `#[cfg_attr(coverage, coverage(...))]`, as well as updating the spots where the feature is enabled.

uglyoldbob:
I would also like to see this implemented.

ToppDev:
Any updates on this?

PR #819, opened 2025-06-20
Guard unstable coverage flags behind independent cfg

complexspaces:
This PR prevents tools like `cargo-llvm-cov` from trying to enable nightly features on stable, while still allowing other downstream crates to use `cfg(coverage)` for other things when coverage tools set the attribute.

I've tested this locally in an internal repo that uses `wayland-client` via `wl-clipboard-rs` and it resolved the nightly feature compilation failures.

Once the coverage attribute stabilizes, and `wayland-rs` bumps its MSRV, this could be removed in favor of just `cfg(coverage)` again.

Resolves https://github.com/Smithay/wayland-rs/issues/717

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -286,7 +286,7 @@ jobs:
           args: --workspace --exclude 'wayland-tests'
         env:
           LLVM_PROFILE_FILE: "wayland-rs-%p-%m.profraw"
-          RUSTFLAGS: "-Cinstrument-coverage --cfg coverage"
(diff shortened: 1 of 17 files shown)


ids1024 (approved):
Just getting to this since I want to do a release soon.

Seems like a good approach. Presumably if anything that depends on `wayland-rs` were to enable a `--cfg unstable_coverage`, it would be for something similar, with a nightly compiler, so it shouldn't cause unexpected problems. Though it will be nice to see the attribute just stabilized so it isn't needed.

PR #821, opened 2025-07-04
protocols: Add experimental

dcz-self:
It comes behind a new feature. Weirdly enough, I can't get cargo expand to generate anything for any of the protocols.

### DIFF
--- a/wayland-protocols-experimental/CHANGELOG.md
+++ b/wayland-protocols-experimental/CHANGELOG.md
@@ -7,3 +7,6 @@
 - Introduce protocol `session_management-v1`,
 - Introduce protocol `input_method-v2`.
 
+### Breaking changes
+
+- Bump `input_method-v2` to version 2.
--- a/wayland-protocols-experimental/Cargo.toml
+++ b/wayland-protocols-experimental/Cargo.toml
@@ -2,7 +2,7 @@
 name = "wayland-protocols-experimental"
 # The major version updates every time the protocols are pulled from wayland-protocols.
 # That's because experimental protocols version bumps are incompatible by definition.
-version = "20250703.0.1"
+version = "20250721.0.1"
 documentation = "https://docs.rs/wayland-protocols-experimental/"
 repository = "https://github.com/smithay/wayland-rs"
 authors = ["Elinor Berger <elinor@safaradeg.net>"]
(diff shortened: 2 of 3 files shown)


ids1024:
Since this won't offer normal stability/semver guarantees, I wonder if a more verbose name than `experimental` would be suitable.

Trying to think of which crates I've seen with a feature flag that enabled apis that don't follow semver... I know some do.

dcz-self:
To be honest, I think it would be better placed in a separate crate, where it can follow semver. But then I looked at the structure of the repo and I saw that there's a preference to keep the protocol files in the same crate directory, so I did this instead.

But I can create a separate crate and pull xml files from "../wayland-protocols/protocols/...". Does this make sense?

Alternatively, I could move the wayland-protocols subrepo from "wayland-protocols/protocols/" and into "protocols/".

ids1024:
I guess the other option would be to just copy the xml file for the experimental protocol into the repository. Maybe better than a duplicate submodule? Or I don't know how `cargo publish` handles symlinks, if that might also work.

From a semver perspective, in theory it would be best to have a separate crate per experimental protcol, with a semver bump whenever a protocol has a breaking change.

dcz-self:
I guess I can get started by simply copying the experimental/directory. Thanks, I'll do that.

dcz-self:
I created a separate crate with copied protocol files, as suggested.

ids1024:
```
error: current package believes it's in a workspace when it's not:
current:   /home/ian/src/wayland-rs/wayland-protocols-experimental/Cargo.toml
workspace: /home/ian/src/wayland-rs/Cargo.toml

this may be fixable by adding `wayland-protocols-experimental` to the `workspace.members` array of the manifest located at: /home/ian/src/wayland-rs/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

Then building with `--all-features`  gets `Cannot find value `XDG_TOPLEVEL_INTERFACE` in this scope`. `[wayland_protocols::xdg::shell]` needs to be passed to the macro.

ids1024:
> Also, no one said that semver major version bump must break all interfaces, right?

Yeah. Also it's possible to depend on two different major versions (but not two minor versions) of a library. So if a smithay for instance had updated to the latest version of one protocol but not the other, it could depend on both versions of wayland-protocols-experimental.

Maybe that's awkward, but after all it's all experimental, so we can see how this works as we try to use it.

ids1024 (approved):

PR #825, opened 2025-07-08
Fix test without system backend

dcz-self:
Without this, `cargo test -p wayland-backend` fails to compile.


ids1024:
Hm. I know there are some issues with `cargo test` with some combinations of features, but `cargo test -p wayland-backend` seems to be working for me on `master`.

dcz-self:
Now it works for me, too. I can't explain it...

dcz-self:
That sounded probable, but I checked and that's still not it.

PR #843, opened 2025-11-10
Add missing protocol version check in rs backend

### LINKED ISSUE #820: wayland-server sends error on calling a method not supported by bound protocol version on sys, but not rs backend
`libwayland` has this code to generate an error:

```c
if (!(resource_flags & WL_MAP_ENTRY_LEGACY) &&
    resource->version > 0 && resource->version < since) {
	wl_resource_post_error(client->display_resource,
			       WL_DISPLAY_ERROR_INVALID_METHOD,
			       "invalid method %d (since %d < %d)"
			       ", object %s#%u",
			       opcode, resource->version, since,
			       object->interface->name,
			       object->id);
	break;
}
```

But without `server_system`, nothing similar seems to be sent by the `rs` backend.

he1senbrg:
Fixes #820 

This patch adds the missing protocol version check to the rs backend so that it sends error on calling a method not supported by bound protocol version similar to the C backend.

### DIFF
--- a/wayland-backend/src/test/protocol_error.rs
+++ b/wayland-backend/src/test/protocol_error.rs
@@ -317,3 +317,65 @@ expand_test!(protocol_error_in_request_without_object_init, {
     // the server should not panic, and gracefull accept that the user did not provide any object data for
     // the already destroyed object
 });
+
+expand_test!(protocol_version_check, {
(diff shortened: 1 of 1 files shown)


he1senbrg:
@ids1024 I've made the changes.

ids1024 (approved):

ids1024:
Thanks!

PR #855, opened 2025-12-30
backend: `set_max_buffer_size`/`set_default_max_buffer_size` methods, on rs backend and libwayland 1.23+

ids1024:
Adds a test and `rs` backend implementation for the function added in https://github.com/Smithay/wayland-rs/pull/815, based on top of https://github.com/Smithay/wayland-rs/pull/851.

I still want to add bindings (and `rs` impls) corresponding to `wl_display_set_max_buffer_size` and `wl_client_set_max_buffer_size`, but that shouldn't be hard. The main thing is adding support to the `rs` backend, and verifying consistent behavior.

We'll need add rounding to the `rs` backend like `get_max_size_bits_for_size` in libwayland. Maybe it would be best to use `NonZeroUsize` in the server and `Option<NonZeroUsize>` in the client,  to express that unbounded buffering is an option, but only in the client...

### DIFF
--- a/wayland-backend/CHANGELOG.md
+++ b/wayland-backend/CHANGELOG.md
@@ -6,6 +6,8 @@
 - client: Added `display_ptr` method to `ObjectId`
 - server: Addded `Handle::global_name`
   * Requires `libwayland_server_1_22` feature
+- client: Added `set_max_buffer_size` method
+- server: Added `set_default_max_buffer_size` and `set_max_buffer_size` methods
 
 ## 0.3.12 -- 2025-12-30
 


ids1024:
Hm, I though it was safe to rely on https://gitlab.freedesktop.org/wayland/wayland/-/commit/d074d52902633d8700ce06b484508db0f8fba02b, since the commit is dated 2021, but it was only merged in 2024.

So like https://github.com/Smithay/wayland-rs/pull/735, we probably want to use https://github.com/elinorbgr/dlib/pull/22.

CoffeeImpliesCode:
It seems like @elinorbgr has left GitHub in 2024 though?

ids1024:
I've released https://crates.io/crates/dlib/0.5.3, and I think I have a solution for how for to deal with libwayland version feature flags (used in https://github.com/Smithay/wayland-rs/pull/873), although there's not a perfect way to handle that and the two backends.

So this should be unblocked now.

ids1024:
Adding a feature flag and testing on CI with an older and newer libwayland version seems to work.

PR #861, opened 2026-01-07
Exclude tests for `backend` and `scanner`

SwishSwushPow:
Hi everyone,
during a dependency review we noticed that the `wayland-backend` and `wayland-scanner` crates include tests in the package published to crates.io. Reducing the amount of code (including the number of test files) makes reviewing crates and the whole supply chain much easier and since you already make use. of `include` in one of your `Cargo.toml` files I thought I'd go ahead and open this PR to apply this to `backend` and `scanner` as well.

Please let me know if anything needs changing or if there are further questions.

### DIFF
--- a/wayland-backend/Cargo.toml
+++ b/wayland-backend/Cargo.toml
@@ -12,6 +12,13 @@ keywords = ["wayland"]
 description = "Low-level bindings to the Wayland protocol"
 readme = "README.md"
 build = "build.rs"
+include = [
+    "README.md",
+    "LICENSE.txt",
+    "CHANGELOG.md",
+    "/src/**/*.rs",
+    "build.rs"
+]
 
 [dependencies]
 wayland-sys = { version = "0.31.8", path = "../wayland-sys", features = [] }
--- a/wayland-scanner/Cargo.toml
+++ b/wayland-scanner/Cargo.toml
@@ -11,6 +11,12 @@ keywords = ["wayland", "codegen"]
 edition = "2021"
 rust-version = "1.65"
 readme = "README.md"
+include = [
+    "README.md",
+    "LICENSE.txt",
+    "CHANGELOG.md",
+    "/src/**/*.rs"
+]
 
 [lib]
 proc-macro = true


SwishSwushPow:
Thank you for taking a look at this. 🙏 

As far as I am aware there are no conventions set in stone in the Rust ecosystem. With the rise of supply chain attacks we are personally convinced that reducing the amount of excess files in a packaged crate can only be a good thing because it makes reviewing dependencies easier and reduces the number of attack vectors. :)

You are correct that removing tests from crates.io makes it harder for crater to access and run these tests. I was looking for a bit more documentation about how crater is utilized by the compiler team, but I couldn't find a whole lot. A colleague pointed out resources about a recent crater run here (https://crater-reports.s3.amazonaws.com/2025-12-16-next-solver-global/index.html) and I had a look at the downloadable results json file.

```
{
  "name": "wayland-scanner-0.31.7",
  "url": "https://crates.io/crates/wayland-scanner/0.31.7",
  "krate": {
    "Registry": {
      "name": "wayland-scanner",
      "version": "0.31.7"
    }
  },
  "status": "",
  "res": "test-pass",
  "runs": [
    {
      "res": "test-pass",
      "log": "master%2361cc47e367d7be91e13bcd01e4e96e0e233d4f6d/reg/wayland-scanner-0.31.7"
    },
    {
      "res": "test-pass",
      "log": "try%23c9557a20e5ef2e0e02c29a5fafd850d1b7cf1e4f/reg/wayland-scanner-0.31.7"
    }
  ]
}
{
  "name": "wayland-backend-0.3.11",
  "url": "https://crates.io/crates/wayland-backend/0.3.11",
  "krate": {
    "Registry": {
      "name": "wayland-backend",
      "version": "0.3.11"
    }
  },
  "status": "",
  "res": "build-fail",
  "runs": [
    {
      "res": "build-fail:compiler-error(E0282, E0425, E0433)",
      "log": "master%2361cc47e367d7be91e13bcd01e4e96e0e233d4f6d/reg/wayland-backend-0.3.11"
    },
    {
      "res": "build-fail:compiler-error(E0282, E0425, E0433)",
      "log": "try%23c9557a20e5ef2e0e02c29a5fafd850d1b7cf1e4f/reg/wayland-backend-0.3.11"
    }
  ]
}
```
So the tests for `wayland-scanner` are actually included, but for `wayland-backend` there seem to be other issues (maybe related to `wayland-sys`).

I also found many other wayland related crates where the source including tests was pulled from github directly, but again, I couldn't learn more about the rational behind that. 😅

It might also be interesting to you that other popular crates such as `reqwest` or `rustls` also dropped the tests (e.g. recently for `reqwest` with some more discussion here: https://github.com/seanmonstar/reqwest/pull/2759).

PR #869, opened 2026-02-13
feat: Add display_ptr() to ObjectId

notgull:
This allows for the display pointer to be determined from the
ObjectId. This is useful for use cases where we need to call
functions on an `ObjectId` in environments where we do not have
the canonical `Display`.


notgull:
This is correct. Otherwise we need to create an entire new `Backend`, which is expensive.

notgull:
Although, this problem could also be solved if `get_data()` were available on `ObjectId` without going through the `Backend`, since the `Backend` is not needed for this call on the library backend.

wayland-backend/src/sys/client_impl/mod.rs
```diff
@@ -146,6 +146,14 @@ impl InnerObjectId {
             std::ptr::null_mut()
         }
     }
+
+    pub fn display_ptr(&self) -> *mut wl_display {
+        if self.alive.as_ref().map(|alive| alive.load(Ordering::Acquire)).unwrap_or(true) {
+            unsafe { ffi_dispatch!(wayland_client_handle(), wl_proxy_get_display, self.ptr) }
+        } else {
+            std::ptr::null_mut()
+        }
```
ids1024: This matches `as_ptr()` above, but I wonder if we should instead be returning a `Result` or `Option`. Returning a `NULL` pointer where the caller doesn't expect it could definitely be bad...

notgull: I'd be fine with a `Result<NonNull<wl_display>, InvalidId>`.

PR #871, opened 2026-02-25
client,server: Add `system` and `dlopen` features to enable `wayland-backend` features

ids1024:
This avoids the need to directly depend on `wayland-backend` to enable these. This seems generally good. There shouldn't be any harm in having these features, with the caveat that we can't use the absence of the feature in the higher-level crate to make assumptions about what backend is in use.

Adding backend related features to `wayland-client`/`wayland-server` also part of my plan for adding feature flags for `libwayland` versions, for things like https://github.com/Smithay/wayland-rs/pull/735 and https://github.com/Smithay/wayland-rs/pull/855:

- Add features like `libwayland_server_1_22` to `wayland-backend` and `wayland-sys`
- Then have a feature like `libwayland_1_22 = ["wayland-backend/libwayland_server_1_22"]` in `wayland-server`
- The functionality in `wayland-server` will be restricted to only apply when the `libwayland_1_22`
- If `system`/`server_system` isn't used, the `libwayland_1_22` feature will still be required, but `libwayland-server` will not actually be needed

The last point is a bit awkward, but it's the best way I can think of the keep the features correctly additive, and also allows a wayland compositor to use new server features with the `rs` backend without requiring a particular version of `libwayland-client` for the `winit` backend of the compositor.

### DIFF
--- a/wayland-client/CHANGELOG.md
+++ b/wayland-client/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add `system` and `dlopen` features, enabling `wayland-backend` features
+
 ## 0.31.11 -- 2025-07-28
 
 - Updated Wayland core protocol to 1.24
--- a/wayland-server/CHANGELOG.md
+++ b/wayland-server/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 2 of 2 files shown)


ids1024:
I don't think the FreeBSD CI error can be related to the changes here...

PR #872, opened 2026-02-25
backend: Return `Result` of `NonNull` in `as_ptr`

ids1024:
Like https://github.com/Smithay/wayland-rs/pull/869, we should probably return a `Result` here; in the next breaking update.


ids1024:
Combined with https://github.com/Smithay/wayland-rs/pull/901.

PR #887, opened 2026-03-24
Support `wl_fixes.ack_global_remove`

Drakulix:
Draft implementation for https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/533.

Based on https://github.com/Smithay/wayland-rs/pull/886.

This needs some feedback for the approach taken. Due to the asynchronous delivery of the global-remove event through `enqueue_event`, we cannot directly call `wl_fixes.ack_global_remove` in the wl_registry handler, as previously enqueued events might still refer to the global in question possibly prompting the client to send requests to it. As such we need to delay the ack until we know the client has had a chance to handle the remove event.

I opted to allow a custom callback function to be pushed into the queue, but I am not sure, if this is the best approach.

Additionally this needs more documentation, that clients *will* receive protocol errors, if they refer to dead global names after the removal event.

### DIFF
--- a/wayland-client/src/event_queue.rs
+++ b/wayland-client/src/event_queue.rs
@@ -338,6 +338,21 @@ impl<State> EventQueueInner<State> {
         I: Proxy + 'static,
     {
         let func = queue_callback::<I, U, State>;
+        self.enqueue_handler(func, msg, odata);
+    }
+
+    pub(crate) fn enqueue_handler(
+        &mut self,
+        func: fn(
+            &Connection,
+            Message<ObjectId, OwnedFd>,
+            &mut State,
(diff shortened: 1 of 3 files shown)


ids1024:
After https://github.com/Smithay/wayland-rs/pull/910 this should be a little simpler; it can just be called in the `Dispatch` impl after invoking `runtime_remove_global`.

Should be simply to update and merge once `libwayland` releases. Or I'd do it now.

ids1024:
This does indeed seem to be quite simple to implement now: https://github.com/Smithay/wayland-rs/pull/944

Though it still needs testing.

Closing this PR in favor of that one.

Issue #801, opened 2025-03-28
client/rs: Using destroyed object in argument still sends request to server, unlike on sys backend


ids1024:
This minimal test case gets an `Unknown id` protocol error with the rs backend, but not with `wayland-backend/client_system`.

Presumably these should be consistent, and the rs backend should also drop the `get_subsurface` call here.

```rust
use wayland_client::{
    delegate_noop,
    globals::{registry_queue_init, GlobalListContents},
    protocol::{wl_compositor, wl_registry, wl_subcompositor, wl_subsurface, wl_surface},
    Connection, Dispatch, QueueHandle,
};

struct State;

fn main() {
    let conn = Connection::connect_to_env().unwrap();
    let (globals, mut event_queue) = registry_queue_init(&conn).unwrap();
    let qh = event_queue.handle();

    let compositor = globals.bind::<wl_compositor::WlCompositor, _, _>(&qh, 1..=1, ()).unwrap();
    let subcompositor =
        globals.bind::<wl_subcompositor::WlSubcompositor, _, _>(&qh, 1..=1, ()).unwrap();

    let surface = compositor.create_surface(&qh, ());
    let surface2 = compositor.create_surface(&qh, ());
    surface.destroy();
    let _subsurface = subcompositor.get_subsurface(&surface, &surface2, &qh, ());

    loop {
        event_queue.blocking_dispatch(&mut State).unwrap();
    }
}

impl Dispatch<wl_registry::WlRegistry, GlobalListContents> for State {
    fn event(
        _: &mut Self,
        _: &wl_registry::WlRegistry,
        _: wl_registry::Event,
        _: &GlobalListContents,
        _: &Connection,
        _: &QueueHandle<State>,
    ) {
    }
}
delegate_noop!(State: wl_compositor::WlCompositor);
delegate_noop!(State: ignore wl_surface::WlSurface);
delegate_noop!(State: wl_subcompositor::WlSubcompositor);
delegate_noop!(State: wl_subsurface::WlSubsurface);
```

PR #894, opened 2026-03-30
backend/rs/client: `InvalidId` on using client_destroyed object in arg, and add test

ids1024:
https://github.com/Smithay/wayland-rs/issues/801


ids1024:
I guess it would also be good to have a should-fail test where we somehow make sure the client sends a message with a destroyed object and, and ensure the compositor errors with that. (Do we have an invalid ID test already?)

PR #897, opened 2026-04-07
WIP Test for destroyed object in non-null argument

ids1024:


ids1024:
Somewhat different case from https://github.com/Smithay/wayland-rs/pull/894.

And I see a different error here than the issue I was trying to reproduce in the test...

PR #898, opened 2026-04-19
client: support complex trait bounds via optional where clauses in macros

Leo-Tinkeam:
We can't simply use `@<T: SomeTrait, U: CustomTrait<T>>` because "<>" are not [“grouping” tokens](https://lukaswirth.dev/tlborm/syntax-extensions/source-analysis.html#token-trees)

There is another solution using "[]" that I have tested and is working but it's a breaking change (or duplicated code)

This solution with where allow us to use `@<T, U> ... where T: SomeTrait, U: CustomTrait<T>` and for those who have already done something like `@<T: 'static + MyTrait> ...`, nothing is broken since the `where` clause is optional

I wanted to use it for `delegate_dispatch` but since `delegate_noop` does almost the same, I added it there to to keep consistency

I did not add it to `event_created_child` as the `lt`, `clt` and `dlt` are not used inside (should be deleted ?) so bounds even if added would not be used either.

### DIFF
--- a/wayland-client/src/event_queue.rs
+++ b/wayland-client/src/event_queue.rs
@@ -795,8 +795,8 @@ impl<I: Proxy, U: std::fmt::Debug, State> std::fmt::Debug for QueueProxyData<I, 
 /// ```
 #[macro_export]
 macro_rules! delegate_dispatch {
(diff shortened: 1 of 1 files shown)


Leo-Tinkeam:
If there is anything else that I should do to be able to merge this, you can tell me

PR #901, opened 2026-04-27
backend: Wrap `wl_interface` to make `wayland-sys` a private dep

ids1024:
We could extract the shared types into yet another crate, or maybe make the re-exports `#[doc(hidden)]`, only to be used by wayland-scanner. Or just copy the definitions here.

But adding `#[repr(transparent)]` wrapper types seems to work, so we may as well do that.

Ideally it would be good to reduce some of the duplication between this and `Interface`/`MessageDesc`. Not sure how much that can be done.

### DIFF
--- a/wayland-backend/CHANGELOG.md
+++ b/wayland-backend/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### Breaking changes
 - Use `Message<_, BorrowedFd>` instead of `RawFd`
-- Wrap `wl_interface` so `wayland-sys` can be a private dependency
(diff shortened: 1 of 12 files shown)


ids1024:
I guess to be a private dependency, we'd also need to change `ObjectId`'s `from_ptr`, `as_ptr` and `display_ptr`.

The main use case for these is with `raw-window-handle` which uses `c_void` anyway.

Also need to consider the `wayland-sys` dependency in `wayland-egl`. That likewise could just use `c_void`.

PR #904, opened 2026-04-28
client,server,scanner: Use default impls in `Proxy`/`Resource`

ids1024:
A couple methods here can be trivially implemented in terms of other methods, so we don't need codegen to generate the same code for each of them.

Adding a default impl of the trait is not a breaking change. I guess this could technically be a breaking change if `wayland-scanner` was updated without updating `wayland-client`/`wayland-backend`? As a proc macro crate, it doesn't specify what version of the runtime dependency it requires...

That's probably fine though.

### DIFF
--- a/wayland-client/src/lib.rs
+++ b/wayland-client/src/lib.rs
@@ -247,7 +247,9 @@ pub trait Proxy: Clone + std::fmt::Debug + Sized {
     }
 
     /// Access the user-data associated with this object
-    fn data<U: Send + Sync + 'static>(&self) -> Option<&U>;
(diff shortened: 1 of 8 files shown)


ids1024:
It looks like `data()` can be implemented in terms of `object_data()` if we add a `'static` bound to the trait. Technically breaking, though in practice these traits are only implemented by `wayland-scanner`, where that is already true.

ids1024:
Or well, that's only an issue in the `wayland-server` version of `data()`. I still need to understand how "variance" works with `PhantomData`...

PR #905, opened 2026-04-30
backend: Don't use `downcast_rs` for `ErasedState`

ids1024:
This doesn't actually simplify anything here.

This is not an API break since `ErasedState` is not public.


ids1024:
I guess this requires Rust 1.86.

PR #908, opened 2026-04-30
client,server,scanner: Eliminate `WEnum`

ids1024:
Extension of https://github.com/Smithay/wayland-rs/pull/793 to also eliminate `WEnum` for enums. Instead, enums generate a tuple struct wrapping a `u32` with associated constants, similar to bitflag enums.


ids1024:
I guess `wayland-scanner` can't actually tell whether the enum used in a method is a bitfield, since wayland-protocols can use enums from the core protocol (etc) but the scanner doesn't have access the that data.

That could be an issue for https://github.com/Smithay/wayland-rs/pull/793, but if we're eliminating `WEnum` completely, it should be addressable.

ids1024:
Oh I almost forgot, but this definitely will need a custom debug implementation (for non-bitflags enums; bitflags already will have that).

PR #909, opened 2026-05-20
backend,egl: Use `NonNull<c_void>` in public API

ids1024:
This, combined with the change previously made in
https://github.com/Smithay/wayland-rs/pull/901, should make wayland-sys a private dependency of wayland-backend.


ids1024:
Also need to make sure `NonNull` vs `*mut` is used consistently in the public API.

PR #910, opened 2026-05-20
client: Update `GlobalList` to use a `GlobalListHandler` trait instead of manual `Dispatch`

ids1024:
If `delegate_dispatch!` is removed, it would be good in smithay-client-toolkit doesn't require a `delegate_registry!` macro to implement `Dispatch` for `GlobalListContents`.

At a minimum, we can simply provide a trait here that is a little bit simpler than `Dispatch`. We know new events won't be added. Also manually implementing `Dispatch` is potentially less clear of an API when that implementation won't get initial globals.

I was wondering if we could then eliminate the use of the lower-level API here (`send_constructor`, cloning and forwarding `Message`), but `registry_queue_init` then can't get initial globals since the `State` isn't available yet.

### DIFF
--- a/wayland-client/CHANGELOG.md
+++ b/wayland-client/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Replace `delegate_noop!` with `Noop` and `NoopIgnore` types providing generic
   dispatch implementations
 - `BindError` now includes the requested and available version, or interface name that failed to bind.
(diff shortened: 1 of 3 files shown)


ids1024:
This seems to be good now, so I'll merge it. If the API isn't perfect it can be adjusted again before release.

We should be able to release wayland-rs soon, though the API changes need some testing in consumers first.

PR #911, opened 2026-05-27
WIP backend: Don't use `downcast-rs` in API

ids1024:
We don't want anything from `downcast-rs` to be in the public API, so it isn't technically a public dependency (although that may not matter much). If we don't need it for much, we may as well drop the dep alltogether.

Without the `ObjectData: AsAny`, it seems the compiler can infer the trait is available, but gets "borrowed data escapes outside of method" errors?

### DIFF
--- a/wayland-backend/src/client_api.rs
+++ b/wayland-backend/src/client_api.rs
@@ -24,7 +24,8 @@ pub use crate::types::client::{InvalidId, NoWaylandLib, WaylandError};
 ///
 /// The methods of this trait will be invoked internally every time a
 /// new object is created to initialize its data.
-pub trait ObjectData: downcast_rs::DowncastSync {
+#[allow(private_bounds)]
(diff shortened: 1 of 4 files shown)


ids1024:
It would be nice if the compiler produced error messages that explained something about what's going on with vtables instead of confusing red herrings...

I wonder if we could just avoid some casts and store the correct traits instead of `Any`. I guess at least with Rust 1.86 upcasting that should be doable.

ids1024:
Or I guess it's using `Arc<dyn Any...>` in those places mainly to erase `D` type generics.

ids1024:
It seems Rust 1.86 casts handle everything except one use? Closing in favor of https://github.com/Smithay/wayland-rs/pull/912, which starts by moving to standard casts where supported. We don't seem to need an `IntoAnyArc` trait.

PR #919, opened 2026-05-30
experimental: add set_frozen request and protocol compatibility to xx-input-method

rano-oss:
Update vendored XML to match the experimental protocol changes: add set_frozen, protocol compat negotiation, supported features advertisement, and version bumps (v1 3->5, manager 3->4). Changed to use zwp_text_input_v3(.2)


ids1024:
Presumably this should only be merged and released here after https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/517 is merged?

rano-oss:
Agreed, made it a draft for now.

PR #920, opened 2026-06-01
backend/client/sys: Guard udata get/set behind state lock

ids1024:
This seems to fix the segfault reproduction on https://github.com/Smithay/wayland-rs/issues/885.

Though there may be other cases where a proxy pointer could be used after being freed. For instance if two threads where racing to destroy the same object? For use in `dispatcher_func` itself, maybe libwayland's internal ref counting with prevent this...

### DIFF
--- a/wayland-backend/CHANGELOG.md
+++ b/wayland-backend/CHANGELOG.md
@@ -13,6 +13,8 @@
 
 #### Bugfixes
 - client/sys: Fix deadlock if `Backend` is used in `ObjectData::destroyed`
+- client/sys: Fix some race conditions accessing udata in a thread while it is
+  being set, or the proxy is being destroyed.
 
 ## 0.3.15 -- 2026-03-30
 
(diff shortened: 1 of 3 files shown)


ids1024:
Presumably `server_system` also needs fixes. Though the C API there is a bit different.

ids1024:
Hm, I guess `InnerObjectId::from_ptr` doesn't have a way to access the lock on the correct `Backend` before calling `wl_proxy_get_user_data`?

In the worst case, I guess this just uses a global static lock.

ids1024:
And maybe just a `Mutex` is better if this lock is always very short-lived.

ids1024:
It could be problematic that `ObjectData::destroyed` is called when we have a `ConnectionData` lock and have already unset the udata (this is already how it behaves). Though it looks like `wayland-client` doesn't even use that method.

Not sure how we'd want to do it differently though. Without causing weirdness if two threads race `destroy`.

ids1024:
Yeah, this will be needed in addition to a lock for the proxy. `destroy_object` in two threads, for conflict.

ids1024:
To make sure proxy pointer is always valid, we either need a lock or ref counting. But I think ref-counting isn't an option because we need to destroy the proxies when the libwayland display is destroyed (or it will have invalid pointers into that). So we can't lazily call `wl_proxy_destroy` in that place, anyway.

ids1024:
Actually, since `libwayland` uses the display mutex for most things, we can just use the state lock here instead of trying to be finer grained than that.

Though it may still be desirable/necessary to improve how proxy pointers are managed at some point.

PR #923, opened 2026-06-03
Add tests for SEGV / null dereference issues

ids1024:
Based on reproduction from https://github.com/Smithay/wayland-rs/issues/885#issuecomment-4588357627 by @idleyoungman using Claude Code. With changes, and addition of two related tests.

`test_thread_destroys` seems to be inconsistent. It works with an unbounded loop, but that's not so useful on CI. (Maybe iterating for a certain period of time would be best for CI.)


ids1024:
Merged as part of https://github.com/Smithay/wayland-rs/pull/920.

PR #928, opened 2026-06-07
Update `wayland-protocols` to 1.49

HigherOrderLogic:
As title.


ids1024:
Merged, with a `CHANGELOG` entry added.

ids1024:
Other wayland-rs crates have breaking API changes merged, but `wayland-protocols` doesn't. So I've bumped and released it.

PR #929, opened 2026-06-09
fix(backend): prevent SEGV/panic when destroy_object is called on the display

### LINKED ISSUE #924: Fix `test_thread_destroy_display` test
I added this along with two other tests in 5a4f39af6d48f762bab5f7e5d54f6ad0fd27884a, but left this one while resolving the other two issues.

This is causing a SEGV with `client_system` and a panic without. The SEGV, at least, shouldn't be possible since the caller isn't using unsafe calls.

tianming-1996:
## Summary

- Fixed the `test_thread_destroy_display` test that was causing SEGV/panic
- Added guards in both `client_sys` and `client_rs` backends to prevent `destroy_object` from being called on the display object (id == 1), returning `InvalidId` instead

## Root Cause

The display object is the root of a Wayland connection and has special handling:

- **`client_sys`** (libwayland backend): The display proxy has no `ProxyUserData`, so `wl_proxy_get_user_data` returns NULL, causing a segfault when `destroy_object_inner` tries to reclaim it
- **`client_rs`** (pure Rust backend): The display uses `DumbObjectData` whose `destroyed()` method contains `unreachable!()`, causing a panic

## Changes

| File | Change |
|------|--------|
| `wayland-backend/src/sys/client_impl/mod.rs` | Added display check in `destroy_object` returning `InvalidId` |
| `wayland-backend/src/rs/client_impl/mod.rs` | Added display check in `destroy_object` returning `InvalidId` |
| `wayland-tests/tests/threads.rs` | Enabled `test_thread_destroy_display` test, now verifies `destroy_object` returns error for display |

## Test Plan

- [x] `cargo test -p wayland-tests test_thread` passes (all 3 thread tests)
- [x] `cargo clippy -p wayland-backend` passes

Fixes #924

### DIFF
--- a/wayland-backend/src/rs/client_impl/mod.rs
+++ b/wayland-backend/src/rs/client_impl/mod.rs
@@ -301,6 +301,14 @@ impl InnerBackend {
     pub fn destroy_object(&self, id: &ObjectId) -> Result<(), InvalidId> {
         let mut guard = self.state.lock_protocol();
         let object = guard.get_object(id.id.clone())?;
+
+        // Do not allow destroying the display object; it uses DumbObjectData whose
+        // destroyed() panics, and the display lifecycle is managed by the connection
+        // itself. Attempting to destroy the display will return Err(InvalidId).
+        if same_interface(id.id.interface, &WL_DISPLAY_INTERFACE) {
+            return Err(InvalidId);
+        }
+
         guard
             .map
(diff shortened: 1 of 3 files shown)


ids1024 (approved):
Seems like a reasonable fix. Thanks for figuring this out!

PR #931, opened 2026-06-19
protocols: Bump version to release this crate

ids1024:
The `master` branch has API breaking changes, but `wayland-protocols` does not. So we can release this crate individually without a backport branch, as long as it's released manually rather than through CI.


ids1024:
I guess the one impact this can have is the bump to `rust-version = "1.86"`. But dependent crates are going to have to bump MSRVs (if not already higher than that) soon anyway.

PR #933, opened 2026-06-22
WIP Update `dlib`

ids1024:
Depends on https://github.com/Smithay/dlib/pull/27.

It would be good to address https://github.com/Smithay/dlib/issues/29.


ids1024:
```
libloading@0.9.0 requires rustc 1.88.0
```

Not sure if we want to bump our Rust MSRV quite that high at the moment...

PR #934, opened 2026-06-22
backend: Split `Message` into `Message` and `OwnedMessage`

ids1024:
Splitting these may help achieve things like https://github.com/Smithay/wayland-rs/pull/799, though due to how `libwayland` works, there's no way on the `sys` backend we can turn a `&str` into nul-terminated string without a clone. And I'm not sure how cleanly we can have generated bindings that don't require that....

Maybe with a trait like https://docs.rs/rustix/latest/rustix/path/trait.Arg.html?

Array arguments are easier to address, though not that common.

### DIFF
--- a/wayland-backend/src/client_api.rs
+++ b/wayland-backend/src/client_api.rs
@@ -11,7 +11,7 @@ use std::{
 #[cfg(doc)]
 use std::io::ErrorKind::WouldBlock;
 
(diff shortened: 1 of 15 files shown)


ids1024:
I've reverted the changes to actually use different types, which cause issues. Those can be attempted separately.

Splitting like this seems overall better regardless.

PR #938, opened 2026-07-02
wayland-scanner: Bump `quick-xml` dependency to `0.41`

MarijnS95:
Two vulnerabilities were just issued for `quick-xml`:

https://rustsec.org/advisories/RUSTSEC-2026-0194.html
https://rustsec.org/advisories/RUSTSEC-2026-0195.html

These were both addressed in the `0.41` release, with no new code incompatibilities in `wayland-scanner`.

### DIFF
--- a/wayland-scanner/CHANGELOG.md
+++ b/wayland-scanner/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Update `quick-xml` to 0.41
+
 #### Breaking changes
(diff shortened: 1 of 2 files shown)


ids1024 (approved):

ids1024:
Thanks!

I guess we could backport these to the current release, though that's annoying given memory exhaustion and run time issues are absolutely not vulnerabilities in wayland-scanner.

(The fact `quick-xml` is always doing semver breaking releases makes me wonder if there are any better xml libraries to use...)

PR #944, opened 2026-07-16
client/globals: Send `ack_global_remove` after handling removal

ids1024:
Succeeds https://github.com/Smithay/wayland-rs/pull/887 now that `wayland` 1.26 has been released. This is now quite simple to implement with the changes to how the wayland-client globals code works since the latest release.

Leaving as a draft since it needs some testing.

### DIFF
--- a/wayland-client/src/globals.rs
+++ b/wayland-client/src/globals.rs
@@ -428,6 +428,11 @@ where
             wl_registry::Event::GlobalRemove { name } => {
                 if let Some(global) = self.remove(name) {
                     state.runtime_remove_global(&globals, conn, qh, &global);
+                    if let Some(fixes) = self.fixes.get() {
+                        if fixes.version() >= 2 {
+                            fixes.ack_global_remove(registry, name);
+                        }
+                    }
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


ids1024:
Since we don't send `ack_global_remove` until after removing the global from the `Mutex<Vec<Global>>`, it should be impossible to bind the global after that using the `GlobalList::bind_*` methods.

It's still possible to do so by directly calling `wl_registry.bind`, but the expectation is that that won't generally be done in the upcoming wayland-rs release. This is different in the current release, where `sctk` has it's own implementation of things in `RegistryState`.

Though normally dynamically added globals would be bound in the `runtime_add_global` handler. This would only be an issue anyway if a client wanted to bind globals in another thread for some reason.

ids1024:
I guess on the server-side, we'll presumably also want a function wrapping `wl_fixes_handle_ack_global_remove`, and an equivalent for the `rs` backend. Unless we can do that just once at a higher level.

I guess libwayland-server also expects that to be used alongside `wl_global_set_withdrawn_listener`.

zzag:
> I guess on the server-side, we'll presumably also want a function wrapping `wl_fixes_handle_ack_global_remove`, and an equivalent for the `rs` backend. Unless we can do that just once at a higher level.
> 
> I guess libwayland-server also expects that to be used alongside `wl_global_set_withdrawn_listener`.

Yes. The expected usage is as follows:

- the compositor calls `wl_global_remove()`
- libwayland-server will call the "withdrawn" callback when all clients have acked the `global_remove` event or the relevant `wl_registry` objects have been destroyed, e.g. because a client disconnected or due to `wl_fixes.destroy_registry`

PR #946, opened 2026-07-17
backend: update debug timestamp format to match libwayland

### LINKED ISSUE #941: Update debug log timestamp format
https://github.com/Smithay/wayland-rs/pull/675 updated the timestamp format for logs printed when environment variable WAYLAND_DEBUG=1 to match the unusual behavior of libwayland (printing fractional milliseconds under mod 2^32 microseconds). This timestamp change should soon be safe to undo because libwayland has merged https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/546 which now prints the fractional part of each second without an offset.

mstoeckl:
Now that libwayland 1.26 is out, the previous timestamp format that was used for compatibility with libwayland can safely be replaced. In this PR I am exactly matching libwayland's new format so that comparison between wayland-rs and libwayland debug logs continues to be simple.

This closes https://github.com/Smithay/wayland-rs/issues/941.

### DIFF
--- a/wayland-backend/src/debug.rs
+++ b/wayland-backend/src/debug.rs
@@ -76,10 +76,14 @@ impl<D: Display> Display for DisplaySlice<'_, D> {
 #[cfg_attr(unstable_coverage, coverage(off))]
 fn print_timestamp() {
     if let Ok(timestamp) = SystemTime::now().duration_since(UNIX_EPOCH) {
-        // NOTE this is all to make timestamps the same with libwayland, so the log doesn't look
-        // out of place when sys tries to log on their own.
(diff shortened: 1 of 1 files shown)


ids1024:
Thanks! I'll need to double check this matches libwayland, but I noticed that in the libwayland release notes. We presumably should indeed match it.

ids1024 (approved):
This seems to match the change in https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/546, and testing on 1.26, the time formatting now seems consistent in `wayland-client` with and without the `system` feature.

Thanks!

PR #948, opened 2026-07-22
Backports for 0.31.x

ids1024:
It seems like a good idea to backport the segfault fix from https://github.com/Smithay/wayland-rs/pull/920, even if we do a 0.32.x release of wayland-client/wayland-server soon.

This also includes the bump of `quick-xml`, and the updated version of the core wayland protocol. Once tests pass I'll merge and manually release these four crates.


ids1024:
I guess CI doesn't run for PRs not to master? I'll just merge. Tests work locally, and these changes are already tested on master.

PR #959, opened 2026-08-20
backend/client/rs: Don't return `WouldBlock` on `ReadEventsGuard::read`

ids1024:
This matches the behavior of the `sys` backend, since `wl_display_read_events` seems to read until `WouldBlock` without returning that as an error.

This is a breaking change in code that uses the `rs` backend and relies on this behavior, but that should be corrected so the code can work with `client_system`.


ids1024:
After reviewing my understanding of `wl_display_prepare_read_queue` and `wl_display_read_events` and verifying the `rs` backend is right, it does seem fine to call `read()` once, and even if other threads are trying to read at the same time, treat a success of the function as `WouldBlock` for the purpose of a tokio `AsyncFd`, etc.

Code on the current release can treat `Ok(0)` from `read()` as the same as `WouldBlock`. The next version will not return `WouldBlock` errors.

Issue #950, opened 2026-08-09
wayland_client::Connection::connect_to_env should be unsafe


mstoeckl:
Any code that relies on a unverifiable program level property to invoke an unsafe API cannot be encapsulated in a "safe" library abstraction.

`Connection::connect_to_env` calls two unsafe functions, each of whose preconditions for sound use are not actually guaranteed. The first is a known issue/has a TODO in the code, but I couldn't find any corresponding github issue to discuss either:

* [`std::env::remove_var`](https://doc.rust-lang.org/std/env/fn.remove_var.html): is unsound in multi-threaded programs on Unix, until the C standard libraries somehow resolve this. (glibc might have a [limited workaround for this](https://sourceware.org/bugzilla/show_bug.cgi?id=15607#c12), but it may be expensive in practice and I don't know whether it is enough for Rust.) Technically I think one can write a safe-ish wrapper over std::env::set_var by panicking if any threads exist.

* `OwnedFd::from_raw_fd` (via FromRawFd trait). The OwnedFd struct makes the promise that safe/sound code cannot close its file descriptor. This can be violated in multi-threaded programs if two threads run `Connection::connect_to_env` in parallel, then they might interleave to both load WAYLAND_SOCKET at the exact same time, both invoke `OwnedFd::from_raw_fd` on the same value, and only then attempt `std::env::remove_var`. If one OwnedFd is then dropped, the other will no longer contain a open file descriptor. In single threaded programs, the specific use of `OwnedFd::from_raw_fd`  here is unsound for a different reason. [`std::env::set_var`](https://doc.rust-lang.org/std/env/fn.set_var.html) is explicitly "sound to call in a single-threaded program". But a single threaded program could open a File, convert to an OwnedFd, invoke `OwnedFd::as_raw_fd` to get its value, `std::env::set_var` to set WAYLAND_SOCKET to have that value, and then run `Connection::connect_to_env`, thereby creating two OwnedFds on the same file descriptor.

(One possible way to keep `Connection::connect_to_env` safe-ish without requiring a very early connection attempt is to split out a separate `unsafe fetch_wayland_socket() -> Option<OwnedFd>` function that gets invoked early by `main` (which _can_ conclusively determine that it hasn't modified WAYLAND_SOCKET or created any file descriptors using its value), and then pass the function's output as a second argument to `Connection::connect_to_env`, which would use it if Some and otherwise connects according to WAYLAND_DISPLAY.)

ids1024:
[...]

PR #961, opened 2026-08-21
client: Mark `connect_to_env` as unsafe, and add seperate `connect_to_env_threadsafe`

ids1024:
One idea I had for https://github.com/Smithay/wayland-rs/issues/950

It's not great to have to do this, but it's also not ideal that `Connection::connect_to_env()` is not really safe despite being marked as such. Nor is this a theoretical issue; it could definitely lead to unsoundness in race conditions in various things using wayland-rs when `WAYLAND_SOCKET` is set. (In the common case where that env var isn't set, the unsafe code simply isn't used.)

I think the big issue with this is that most applications using wayland-rs aren't actually calling `connect_to_env` directly, but instead using `winit`, or a toolkit or game engine (that may or may not use `winit` internally).

I don't expect those libraries to implement strict rules about only being called at the start of the main thread. So in practice does that mean they just use `unsafe { }` and still have soundness issues, or just use `connect_to_env_safe()`?

The latter is **probably** fine? But breaks the expectation of basically every other Wayland client, which will use `WAYLAND_SOCKET` if present. I don't know how much `WAYLAND_SOCKET` is used in practice.

Maybe we should also have a sanity check that the FD in `WAYLAND_SOCKET` isn't 0-2. Can we assume at the start of `main()` no FD should be opened by the process? If so we could mostly be safe by checking that `WAYLAND_SOCKET` is an open FD and isn't 0, 1, or 2.

# Other Options

* We could just leave it as is and hope there aren't issues. But I think this is risky particularly when Rust code ends up using multiple libraries that try to create Wayland connections. If those happen on different threads, they are potentially racing with each other (and other things) on the env var soundness issue, but also could both end up trying to use the same `WAYLAND_SOCKET` FD.

  Multiple Wayland connections in one client is probably undesirable, but should be sound. And seems to happen sometimes in Rust applications in practice. There is also no way `WAYLAND_SOCKET` will work as expected here, since only one instance can actually use it.

* We could try to check if other threads exist and panic or return an error if any do. Maybe that would be good to add to the current version as an extra check, but it doesn't address the possibility `WAYLAND_SOCKET` is just bogus.

* We could avoid the whole thing by just unconditionally ignoring `WAYLAND_SOCKET`. Again, it probably isn't used often, but other Wayland clients use it.


ids1024:
Winit already makes building an event loop off the main thread panic unless `EventLoopBuilderExtWayland::with_any_thread` is used. That doesn't strictly guarantee it will meet the safety guarantees (there could be other threads at that time, and `with_any_thread` isn't unsafe) but it can probably continue to use the unsafe `connect_to_env`.

ids1024:
https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/553 adds documentation about the thread-safety to `libwayland`.

wayland-client/src/conn.rs
```diff
@@ -39,56 +39,89 @@ pub struct Connection {
...
+    };
+
+    UnixStream::connect(socket_path).ok()
+}
+
 impl Connection {
     /// Try to connect to the Wayland server following the environment
     ///
+    /// This first attempt to connect to the file descriptor specified in `WAYLAND_SOCKET`.
+    /// And then removes that file descriptor from the environment. If that variable does
```
mstoeckl, line 91: Nit: this removes the environment variable, not the file descriptor; although it does set CLOEXEC and prevent further inheritance.

wayland-client/src/conn.rs
```diff
@@ -39,56 +39,89 @@ pub struct Connection {
     pub(crate) backend: Backend,
 }
 
+unsafe fn stream_from_wayland_socket_var() -> Result<Option<UnixStream>, ConnectError> {
+    if let Ok(txt) = env::var("WAYLAND_SOCKET") {
+        // We should connect to the provided WAYLAND_SOCKET
+        let fd = txt.parse::<i32>().map_err(|_| ConnectError::InvalidFd)?;
+        let fd = unsafe { OwnedFd::from_raw_fd(fd) };
```
mstoeckl, line 46: This will panic if WAYLAND_SOCKET=-1. (Since file descriptor values are always nonnegative, this it is safe to just filter out all negative values.)

wayland-client/examples/list_globals.rs
```diff
@@ -36,7 +36,7 @@ impl Dispatch<wl_registry::WlRegistry, AppData> for () {
 fn main() {
     // Create a Wayland connection by connecting to the server through the
     // environment-provided configuration.
-    let conn = Connection::connect_to_env().unwrap();
+    let conn = unsafe { Connection::connect_to_env() }.unwrap();
```
mstoeckl, line 39: While it may be quite tedious to write, I think adding a [SAFETY comment](https://std-dev-guide.rust-lang.org/policy/safety-comments.html#safety-comments) could be helpful to explain why this location of the call is safe (e.g. "Called immediately in main()"). This should reduce the risk that otherwise people building directly off the examples may refactor the code they have copy-pasted and refactor away the preconditions for `connect_to_env()` to be used.

wayland-client/src/conn.rs
```diff
@@ -39,56 +39,89 @@ pub struct Connection {
...
+        // remove the variable so any child processes don't see it
+        // TODO: Audit that the environment access only happens in single-threaded code.
+        unsafe { env::remove_var("WAYLAND_SOCKET") };
+        // set the CLOEXEC flag on this FD
+        let flags = rustix::io::fcntl_getfd(&fd);
+        let result = flags
+            .map(|f| f | rustix::io::FdFlags::CLOEXEC)
+            .and_then(|f| rustix::io::fcntl_setfd(&fd, f));
```
mstoeckl, line 54: > Maybe we should also have a sanity check that the FD in WAYLAND_SOCKET isn't 0-2. Can we assume at the start of main() no FD should be opened by the process? If so we could mostly be safe by checking that WAYLAND_SOCKET is an open FD and isn't 0, 1, or 2.

I think such a safety check would indeed be worth it and is unlikely to break programs, because even doubly-forked daemons tend to get a `dup2`'d /dev/null for 0,1,2 instead of using them for inherited file descriptors.

In addition: the file descriptor from WAYLAND_SOCKET necessarily cannot have had CLOEXEC set to begin with, or this process could not have received it. Erroring if this property does not hold could help reduce the risk of capturing the wrong file descriptor. (Example: a program opens a config file with file descriptor X before calling connect_to_env and then WAYLAND_SOCKET by chance had value X; well written programs and APIs should create X as CLOEXEC; I believe Rust's standard library consistently does this.)

wayland-client/src/conn.rs
```diff
@@ -39,56 +39,89 @@ pub struct Connection {
...
+        match result {
+            Ok(_) => {
+                // setting the O_CLOEXEC worked
+                Ok(Some(UnixStream::from(fd)))
+            }
+            Err(_) => {
+                // something went wrong in F_GETFD or F_SETFD
+                Err(ConnectError::InvalidFd)
```
mstoeckl, line 62: If `fcntl_getfd` returned `EBADF`, then the OwnedFd Drop implementation will call close on an invalid file descriptor. While OwnedFd currently ignores _all_ errors because EINTR and EIO are possible thanks to some filesystem misdesigns, I believe it could choose to panic if on EBADF in the future.

mstoeckl (commented):
I've added a few comments, some which are arguably on preexisting issues. I don't think there will be any "good" solution here until major libc implementations offer threadsafe env manipulation, and Rust detects that; this may take another few decades. So this might be the best option.

wayland-client/src/conn.rs
ids1024: Yeah, we can probably assume 0, 1, and 2 are open and used for stdin/stdout/stderr, or placeholders for those. While technically not required by the OS, that's basically required in normal applications.

> Erroring if this property does not hold could help reduce the risk of capturing the wrong file descriptor.

Oh yeah, checking for `CLOEXEC` could be another good sanity check to add here. It doesn't make it thread-safe since these things aren't atomic (along with the inherent unsafety of `unsetenv`) but it is probably a fair check to have.

We could also do propose this to `libwayland`.

wayland-client/src/conn.rs
ids1024: Good catch. We should definitely return `InvalidFd` for negative values.

wayland-client/src/conn.rs
ids1024: Yeah, I guess Rust could change `OwnedFd::drop` to panic on `EBADF`, even if it basically has to silently ignore any other error from `close()`.

May as well make sure we avoid this.

ids1024:
https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/557 proposes a way to ignore `WAYLAND_SOCKET` to libwayland, now that my documentation merge request (https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/553) was merged.

We can do this in wayland-rs even if the libwayland proposal is rejected or stalled, but we can see what response that gets.

PR #967, opened 2026-09-05
sys: allow forcing no `dlopen` usage

### LINKED ISSUE #930: Cargo feature to force no `dlopen`.
Hi,

Currently I'm writing a GUI app, which only plans support for Wayland but not X11. However, GUI frameworks (and other underneath libraries such as `winit`) enable the `dlopen` feature by default so that X11 user does not need the `libwayland` package installed. Since my app does not support X11, I think this is redundant and would prefer static linking instead.

However, since Cargo features are additive, I cant turn off the `dlopen` feature from my `Cargo.toml`. So I want to propose something like `force-no-dlopen` so that user can toggle this from their side. If you want, I can try to give this a go and create a PR myself.

HigherOrderLogic:
Closes #930.


wayland-sys/build.rs
```diff
@@ -1,7 +1,13 @@
 use pkg_config::Config;
 
 fn main() {
-    if std::env::var_os("CARGO_FEATURE_DLOPEN").is_some() {
+    println!("cargo:rustc-check-cfg=cfg(dlopen)");
+    println!("cargo:rerun-if-env-changed=FORCE_NO_DLOPEN");
+
+    if std::env::var_os("CARGO_FEATURE_DLOPEN").is_some()
+        && std::env::var("FORCE_NO_DLOPEN").map_or(true, |v| v != "1")
```
nekopsykose, line 8: The issue (#930) talks about a cargo feature, but here this is a pure envvar (doesn't start with `CARGO_FEATURE_`). This means it can't be set via features= in Cargo.toml and needs to always exist in env for a given build. Are you sure this is correct? Personally I would rather use a feature for this kind of functionality if I wanted to use it myself.

HigherOrderLogic: Yes, this is intentional. 

As I said, many crates turn on the `dlopen` feature unconditionally, and since features are additional, users are forced to use it. And since creating the issue, I have changed my mind and think that an env var is better, since it would prevent another default features situation like the current one.

nekopsykose: Fair enough, but this really sucks to use, since it's invisible unless one goes through some docs that tells them there's a magic env var that overrides a feature. Except here there aren't any, so one has to read the build.rs itself to discover this. It would still work for me I guess.

HigherOrderLogic: Yeah, agree it's a good idea to add some docs here. Will do later.

PR #972, opened 2026-09-18
backend: Use `&ObjectId` instead of `ObjectId`

ids1024:
This involves reference counting (depending on if it's sys/rs), so it's best to not clone everywhere unnecessarily.


ids1024:
This could potentially also allow `Message`/`Argument` to use `&ObjectId`. Which may be nice if we can get it to use `&[u8]` and `&CStr` too.

PR #973, opened 2026-09-18
backend: Use `&` for remaining uses of `ObjectId`/`GlobalId`/`ClientId`

ids1024:
Continues work started in https://github.com/Smithay/wayland-rs/pull/972.

Since reference counting is used on the `sys` backend, types ID types can't be `Copy`. So methods should accept them by reference, rather than by value. If just to be idiomatic in avoiding clones, if not for any real performance difference.

Maybe the traits though should pass owned versions to methods, if wayland-client and wayland-server need owned versions to wrap them in other types anyway.

Otherwise, this seems clearly better.


ids1024:
Actually, using references in the `wayland-backend` traits may be better overall if we could make wrapper types `#[repr(transparent)]` and use casts. Though I don't think there's currently a way to do those casts without `unsafe`, an extra crate, or both...

Well, the sys backend may as well get an extra use of unsafe, but for `rs` I think it's nice to have as few uses of `unsafe` as possible.

